PR #4, opened 2015-04-02
Implemented targetVolume 8

ryonamin:
This includes a config file modification by Gabrielle.

### DIFF
--- a/config/stdinclude/Materials/TEDD_PS_SupportPlate_updatedPS
+++ b/config/stdinclude/Materials/TEDD_PS_SupportPlate_updatedPS
@@ -1,0 +1,23 @@
+// Endcap disks material
+
+//  #include <generic/CoolingBlocks> // TODO: check this - duplicate entry in ModuleMechanics
+
+Materials TEDD_PS_SupportPlate {
+  type layer
+
+  Component {
+    componentName "PS SupportPlate"
+    service false
+    scaleOnSensor 0
+    Element {
+      elementName CFRP
+      quantity 0.305
+      unit mm
+    }
+   Element {
+      elementName Epoxy
+      quantity 0.057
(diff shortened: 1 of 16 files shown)


alkemyst:
Dear Ryo,
  (sorry for the double message, I am slowly getting used to gitHub)
first of all thank you very much for this addition. I actually have a couple of comments regarding this pull request on possible improvements.
1) the object we are talking about is called "support plate". It would be nice to change the name from bottomHybrid to something generic, like 'bottomVolume' or specific like 'supportPlate' in the code (as you like).
2) as for the thickness, I think it would be more appropriate to define a property named supportPlateThickness in the module, and the access it, just like all of the other mechanical measurements in the module up to now (size of hybrids, size of the sensor, sensor spacing, sensor thickness, etc...).
3) the support plate might or might not be there, depending on the module type. Would it be possible to place it or skip it, based on whether any material has a targetVolume 8 or not? Or anternatively on whether the sensorPlateThickness parameter (described in point 2) is zero or not?

Please let me know what you think of this and whether you need any help/additional info.

Thank you very much
Stefano

ryonamin:
Dear Stefano,
First of all, you have a nice portrait :-)
Thank you for the comments.

> 1)
Yes, I agree. I would take 'supportPlate', which is more informative name.

> 2)
Okay, I will try to implement it, too.

> 3)
Currently if you don't specify any material to targetVolume 8 in your conf file, you will not have the 'supportPlate' volume in your output xml.
Do you think this is good enough?

So I will improve the above points, please reject this PR. I will make another PR later.

Best regards,
Ryo

alkemyst:
Thanks a lot. Again please let me know if you need any help with this!

Cheers
Stefano

PR #5, opened 2015-04-08
Add support plate structure (targetVolume8)

ryonamin:
The previous "HybridVolumes" was renamed as "ModuleComplex" because now it handles not only hybrid volumes but also a support plate.

The config file for targetVolume made by Gabrielle is also included.

Additional fix : I found that there was a naming conflict e.g.
supportR920Z1180 were defined twice.


alkemyst:
Dear Ryo,
  thanks again for this improvement. I have just an additional question (sorry!): would it be possible to exclude the configuration files from this PR? There is some work in progress there and I would not like to mix some well finished and good code development with a half-cooked configuration file...

Thanks again
Stefano

ryonamin:
Dear Stefano,
Sure, no problem! I will make a new PR without config files.
Please close this request.
Best regards,
Ryo

alkemyst:
Thanks a lot.

PR #7, opened 2015-04-09
Corrected bug in material routing boundaries ordering

stefano-martina:


stefano-martina:
I make a pull request with this changes and others

PR #8, opened 2015-04-10
Corrected bug in material routing boundaries ordering and fixed the material routing in case of  barrel modules with maxZ greather than the one of the layers

stefano-martina:


alkemyst:
One of these commit is not working as axpected. I create another pull request, cherry-picking the tested patch https://github.com/stefano-martina/tkLayout/commit/0bc767c04e4d596bca2acd38fb30ad62c0157818

PR #27, opened 2015-04-29
PS material update and reorganization of cfg files

ghugo83:


alkemyst:
Accepting PR30 instead

PR #32, opened 2015-05-04
Add warning messages when mass-less volumes are found and skipped.

ryonamin:
I have changed the way of checking if volumes are empty or not.
Previously it was based on material density but now it is based on material list.


alkemyst:
I agree this is more solid solution. Can I ask you to change the PR to tkLayout:dev instead of tkLayout:master?

Thanks
Stefano

ryonamin:
okay, I will close this PR and make a new one.

PR #33, opened 2015-05-04
Add warning messages when mass-less volumes are found and skipped.

ryonamin:
I have changed the way of checking if volumes are empty or not.
Previously it was based on material density but now it is based on material list.


alkemyst:
Perfectly fine by me! Thanks :+1:

Issue #35, opened 2015-05-06
Cleaning up obsolete material budget code


alkemyst:
Some objects and chuks of code to treat old material configuration files should be removed.
Also, all references to the "exiting" material should be removed: material in modules is now only "local" and the routing of materials in the service volumes is taken care of by the new Stefano Martina's code (materialway and the like)

PR #36, opened 2015-05-06
Dev remove exiting masses

alkemyst:
This PR solves #35


alkemyst:
This was tested and gives exactly the same material amount and resolution as revision dev-697 on https://github.com/tkLayout/tkLayout.git

alkemyst:
Wrong PR: should go to dev instead of master

PR #47, opened 2015-06-10
DevCMake - added cross-platform compiling system with updated README.md

drasal:
Added cross-platform compiling system with updated README.md file. It compiles currently on Debian Linux system & SL6. Change and source setupCMake.sh file according to your system...


alkemyst:
This PR cannot be automatically merged, so I resolved the conflict manually and created the corresponding branch drasal-devCMake.
This will make its way into dev through pull request #48

PR #52, opened 2015-06-17
Updated mechanics material for 2S and PS rods and modules (Baseline2015 and Baseline2015_tilted)

ghugo83:


alkemyst:
Hi,
  after a bit of thought we deemed it better to use @include statements instead of repeating the same information. Could you please move blocks of code like these
```
  // Support Profiles
  // Mechanics
  Component {
    componentName Mechanics
    service false
    scaleOnSensor 0
    targetVolume 0
    Element {
      elementName CFRP
      quantity 1.25
      unit g
    }
  }
```
into dedicated files and then use @include ?

Thanks

PR #61, opened 2015-08-04
Adding the latest moduleOrientation

alkemyst:


alkemyst:
There are some other minor changes in my private dev, which I would like to propagate upstream

PR #63, opened 2015-08-05
Fixing  bug to build correctly wedge-shaped sensors in disc region 

drasal:
Several bugs fixed:

- wedge-shaped polygons were flipped
- a small bug appeared when positioning sensors in discs (not correctly calculated optimal position to have all discs of the same layout taking into account either zError or rOverlap)
- cutting algorithm applied at upper radius limit when building sensors in discs from bottom up doesn't work -> problem when non-zero overlap is required -> basically the radius is never achieved due to non-zero overlap which is always applied when building the next ring, so algorithm usually adds several very small rings to make the cut -> currently commented out, but in the future better algorithm has to be applied

### DIFF
--- a/include/Disk.h
+++ b/include/Disk.h
@@ -36,8 +36,8 @@ private:
   Property<double, NoDefault> innerRadius;
   Property<double, NoDefault> outerRadius;
   Property<double, NoDefault> bigDelta;
-  Property<double, Default> rOverlap;
-  Property<int, Default> bigParity;
+  Property<double, Default>   rOverlap;
+  Property<int   , Default>   bigParity;
 
   PropertyNode<int> ringNode;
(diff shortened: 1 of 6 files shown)


alkemyst:
Dear Zbynek,
  thank you very much for this PR. I've gone through all the new code except ring.cpp (there is something complex there, and I need a little more time to understand that). Anyway: I've seen a possible bug: in disk.cpp line 51 there is (I think) a relic of debugging:
  zError(10);
If this is correct, could you please remove this line and reiterate the PR?
  As I'm there, I am not sure I understood the changes in GeometricModule.cpp (those marked with the following comments:
// Doesn't work as expected -> creates loop to exactly catch the 
// edge (problem when shifting new layer by overlap ...)
  could I please ask you to comment on how will this affect the code?

Thanks
Stefano

PR #98, opened 2015-10-27
Added Units constants

drasal:
Use units as:

double energy =  10 * Units::GeV;

std::cout << "Energy = " << energy / Units::MeV << std::endl


alkemyst:
Hi, could you please add on the fly also um (1e-3) ?
Thanks

drasal:
Micrometers added. Pull request closed and a new one with updated file opened.

PR #99, opened 2015-10-27
Added Units constants

drasal:
Use units as follows:

double energy = 10 * Units::GeV;

std::cout << "Energy = " << energy / Units::MeV << std::endl


drasal:
Unit for micrometers added.

PR #105, opened 2015-11-06
Dev p mom - Added simulation of tracker(s) resolution for given total momenta

drasal:
Added simulation of tracker(s) resolution for given total momenta (not just pT), visualisation adapted accordingly, arranged full layout to be positioned at the top of Index web page.

Apply after #104

### DIFF
--- a/include/Analyzer.h
+++ b/include/Analyzer.h
@@ -364,11 +364,16 @@ namespace insur {
     virtual Material findHitsInactiveSurfaces(std::vector<InactiveElement>& elements, double eta, double theta,
                                               Track& t, bool isPixel = false);
 
-    void clearGraphs(int graphAttributes, const std::string& aTag);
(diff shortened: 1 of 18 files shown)


alkemyst:
After applying the pull requests 101 to 105 the resolution plots with 3000 sample tracks are rebinned a bit too much (see picture). The see-saw part of the red plot representing resolution at the edge between one disk and the next one is washed out. Is there a way to control the rebin?


screenshot from 2015-11-10 17 36 49

drasal:
One can change the visualization of binning by setting the global constant:

 static const double vis_eta_step = 0.1;
 static const int    vis_n_bins = geom_max_eta_coverage/vis_eta_step;

Another way would be to relate the binning to the number of tracks or the eta_step could be defined in a SimParms file. What do you think?

PR #120, opened 2015-11-30
Important : Rings can be placed at different z within a disk

ghugo83:


ghugo83:
Modification of the 'testing' cfg files, for Baseline2015 directory only : 
Reduced hybrid sizes for rings 8 and 10 only.

PR #135, opened 2015-12-10
typo

ghugo83:


alkemyst:
Hi @ghugo83 ,
  could you also add "Sensor" as one of the possibilities?

Thanks

ghugo83:
Yes, sure. Sorry, I should have thought of it ;)
Please note that it was not possible/risky in the code to use sth like : "does the component name contain 'Sensor' ", because there are components like 'Sensor spacers'.

PR #146, opened 2016-01-06
Wrong tracks considered for Track::pruneHits() in src/Analyzer.cc

ghugo83:


drasal:
Hello Gabrielle and Stefano,

I think that the solution with continue command is not correct, one should use curl brackets inside. This would stop program from checking other options. Please, have a look here:

https://github.com/tkLayout/tkLayout/compare/dev...drasal:devPrune

I would create directly pull request, unfortunately I have also mistakenly committed the changes with solution to phi issue. That's still waiting for approval, thus I don't want to ask for a pull request

ghugo83:
Hi Zbynek,
Why don't you just create a branch from upstream/dev and do a 'curly bracket' commit ? That way you can of course have several independant PR at the same time. Or maybe you need the solution to phi issue for this commit ?
Cheers,
Gabrielle

drasal:
Hi Gabrielle,

sorry, I still a little bit struggle with git. I will do it...

  Zb.

PR #147, opened 2016-01-07
Work on parametrized spatial resolution for modules

ghugo83:
- now possible to set either numStripAcross, either pitchEstimate
- now possible to set either numSegments, either stripLengthEstimate
- Resolution estimates put in 'modules' table on the geometry tab on the website
- Improved resolutionLocalX and Y calculations 
- Plots automatatically put on the website
- Option -R (or --debug-resolution)


alkemyst:
Reviewing code with @ghugo83 , we found some possible improvements:
* change the rounding floor(parent_->meanWidth() / pitchEstimate()); into floor(parent_->meanWidth() / pitchEstimate() + 0.5);
* use pitchEstimate/... by default where useful (and remove old numStripsAcross and the like)
* document (in user manual wiki) pitchEstimate/...
* document the possibility of using pitchEstimate + size in the referenceSensor too
* tagMapAveRphiResolution should be an accumulator (accumulating accumulators from each module) -- may need an ad hoc object or not
* change "*1e3" into "/Units::micrometer" (Vizard.cpp:1531)

[ Beautiful code :-) ]

alkemyst:
After inspecting the proposed output page, here's a few more desiderata:
* (if possible) please add histogram statistics in the resolution distribution (right plot only)
* correct the range of ctg(alpha)
* move the plots to the geometry tab

ghugo83:
Done :)

PR #153, opened 2016-01-18
Solved bug in number of hits plot : override setup() function for End…

ghugo83:
…cap modules


alkemyst:
Gabrielle says it is tested.

ghugo83:
Yes indeed ;)
I don't guarantee there is no bug, but at least none that I can see :)

PR #157, opened 2016-01-20
New include system

alkemyst:
Implementing a new @include system (see commit 1892fdd for details)
Better web output for the include tree (with the option -w)


alkemyst:
PR tested and it's fine

PR #158, opened 2016-01-21
Creating new layouts and moving bits and pieces around

alkemyst:
to make the configuration more flexible.
Tested: no regression on Baseline_tilted_200


alkemyst:
Tested on Baseline_tilted_200

PR #160, opened 2016-01-21
solved little bug with -nan in modules tab

ghugo83:


alkemyst:
Hi Gabrielle,
  I might be wrong, but I have the impression that fabs() is covering a problem there. If I understood things properly, one should never get a negative number under square root , by construction...

ghugo83:
Hi Stefano,
Yep sure, but it turned out that (N*Q - S^2) / N^2      (used to approximate sigma^2)
with the notation 
N : count
S : sum
Q : sum of squares
can be negative when it aproximates 0 (I got sth like a -10e(-19) instead of 0).
So the abs() avoids this problem.
Of course when (N*Q - S^2) / N^2  approximates a non-null number, i always get sth positive.
Cheers,
Gabrielle

PR #161, opened 2016-01-22
Dev phi issue - problem with min & max phi calculation when crossing +pi/-pi line.

drasal:
Implemented new algorithm to find if the module was possibly hit to speed-up the whole procedure of tracking algorithm. The couldHit function works correctly only for the rectangular modules, so for wedge-shaped modules it returns always true (correct, but no speed-up effect).


drasal:
Kindly tested by Gabrielle using the baseline geometry -> OK:
http://ghugo.web.cern.ch/ghugo/layouts/resolution/Baseline_flat/indexpixel.html

Compare with the older version:
http://cms-tklayout.web.cern.ch/cms-tklayout/layouts/repository-git-dev/Baseline_flat/indexpixel.html

PR #168, opened 2016-02-12
Geometry files for pixels too

### LINKED ISSUE #96: Csv summary & geometry output files created just for outer tracker, no output for pixel tracker
The Csv summary output is hard-coded for trigger only, pixel output summary not supported. The idea is to extend it for any number of trackers defined. An intermediate step for future object reorganization, having analyzer & vizard connected with a given tracker.

alkemyst:
Fixes #96 

UNTESTED - please test before merging

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
 bin
 lib
 tkgeometryrc
-test/testGraphVizCreator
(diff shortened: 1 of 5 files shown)


alkemyst:
Tested, OK

PR #169, opened 2016-03-07
Removed redundant info in old PIXEL XML template file + When generating the xmls, a cfg info file is created

ghugo83:
- Removed redundant info in old PIXEL XML template file (trackerVolumeTemplate.xml)
- When generating the xmls, a cfg info file is created : the info file is called layout_date.cfg and is placed in the xml directory, in addition to the set of XMLs


alkemyst:
Thank you very much for this, Gabrielle!
Did you also test this against regression?

ghugo83:
Yes I did. 
The only thing I would still want to modify is that the 'comment' part is placed a bit lower in tracker.xml than in the other XMLs, it totally doesn't matter of course since it is just a comment. But I will make things be perfectly 'symmetric' in a few mins and push it ;)

ghugo83:
Done ;)

PR #170, opened 2016-03-11
Several bugs in Vizard::createXmlSite solved + added the metadata file to the XML tab

ghugo83:
Worked on Vizard::createXmlSite. To sum up, the following bugs were corrected : 
- in case of run with --xml, the copy of XML files did not work
- in case new TKG_LAYOUTDIRECTORY, the copy of XML files did not work
- in case run with --xml but no --pixelxml, Warning: RootWBinaryFileList::dump() appeared 
- RootWBinaryFileList objects were not instanciated correctly
- boost::filesystem::copy_file code was replicated
- error handling was incomplete

Also : added the metadata file to the XML tab.

### DIFF
--- a/src/Squid.cc
+++ b/src/Squid.cc
@@ -748,7 +748,8 @@ namespace insur {
    
   void Squid::createAdditionalXmlSite(std::string xmlout) {
     std::string xmlpath = mainConfiguration.getXmlDirectory() + "/" + (xmlout.empty() ? baseName_ : xmlout) + "/";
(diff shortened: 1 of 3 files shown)


alkemyst:
Gabrielle says it's tested, so I will approve it :-)

ghugo83:
Yes. Please note I have also noticed that there is a seg fault with --pixelxml when running tkLayout (whether or not merging this PR, the problem was already there). 
As the pixel XML export is in current development anyway, I will contact the pixel XML export team about it.

PR #176, opened 2016-04-12
URGENT: fixing an integration bug

alkemyst:


alkemyst:
Due to a mistake in the integration  of branches (my fault) the dev compilation was broken. This commit fixes the problem

Issue #89, opened 2015-10-06
Coordinate file for pixel missing


alkemyst:

alkemyst:
Apparently the fix for this did not make its way to the dev branch

alkemyst:
Should be fixed in https://github.com/alkemyst/tkLayout/commit/765820e82b2c62cc9062117fb0122be20eb3bba3 (commit 765820e8)

PR #179, opened 2016-04-21
Cherry-pick from Stefano : Each tracker has its geometry files now.

ghugo83:
+ Moving geometry files to the geometry section.


alkemyst:
This fixes issue #89

PR #182, opened 2016-04-29
Fixing the name of the full layout image

alkemyst:


alkemyst:
Tested locally by Stefano

PR #187, opened 2016-06-16
Fixed MaterialBudget analysis ...

drasal:
* Fix to hadron efficiency plots -> due to missing beam-pipe effect
* Changed way of filling histograms -> eta generated such as it corresponds to histogram bin centers to avoid rounding error (other solution would be usage of Profile histogram)


alkemyst:
Dear Zbynek,
  thanks a lot for this pull request. Actually fixing eta co the bin centres does not correspond to what we would like. Indeed the correct way of dealing with such plots is using TProfiles from the start.

alkemyst:
(i.e. getting rid of TGraphs altogether)

PR #189, opened 2016-06-29
PLEASE CHECK : New parameters values from Ernesto for the module reso…

ghugo83:
…lution parametrization


alkemyst:
Checked and it makes sense

PR #201, opened 2016-07-17
New plots with number of hit layers

alkemyst:
Both in the geometry page of each tracker and the info page


alkemyst:
Tested by Stefano against regression.

PR #202, opened 2016-07-18
Yet another TBPS version

alkemyst:
Plus a tiny fix: removing a debug line which was forgotten in the code


alkemyst:
Plus some indent fixing.
Checked by Stefano against regressions

PR #204, opened 2016-07-18
Merged latest local branch

ghugo83:


alkemyst:
Two comments:
1) the problem is there and should be fixed for 3.6.2, but also for the whole 3.6.X series
2) remoding the rings summary would result in a regression for the outer tracker

PR #206, opened 2016-07-18
Dev lite - Track resolution analysis finished

drasal:
* Analysis of track resolution finished (still missing csv output -> will be implemented soon)
* All color definitions & functionality moved to Tools/Palette singleton class
* Rootweb environment adapted, such as a user can retrieve its content: canvases, histograms, etc 
* Web priority of individual web pages automated

### DIFF
--- a/source/Analysis/include/AnalyzerMatBudget.h
+++ b/source/Analysis/include/AnalyzerMatBudget.h
@@ -113,8 +113,8 @@ class AnalyzerMatBudget : public AnalyzerUnit {
 
 }; // Class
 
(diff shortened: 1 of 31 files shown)


alkemyst:
Hi,
Thanks for the pull request. Commenting here just to mention that the csv output is most probably not needed, as it was developed to be able to compare a large number of similar options.

It can be added if and when needed.

drasal:
I needed to have a look on the csv output as well to test its functionality within the new design. So, don't be surprised by a new pull request :-)

PR #212, opened 2016-07-20
No more warnings and smaller caps

alkemyst:


alkemyst:
New code behaviour tested by Stefano

PR #217, opened 2016-07-27
Adding a geometry with a single (full) TEDD

alkemyst:


alkemyst:
Tested by Stefano:
TEDD_Full_V351_200.cfg is compliant with (TEDD1_V351_200.cfg + TEDD2_V351_200.cfg)

PR #221, opened 2016-08-03
Delphize can now run also on summary.root

alkemyst:


alkemyst:
Tested (also against regression) by Stefano

PR #222, opened 2016-08-08
Added geometry info in csv files on website

ghugo83:
Added in geometry csv files in the website : 
- modules tilt angles;
- modules width and length;
- sensorThickness.

Also improved display of info in csv files.


alkemyst:
Tested against regression by Gabrielle

PR #224, opened 2016-08-11
Added possibility of having different modules sizes in flat part with respect to tilted part (tilted placement) + FIXED BUG IN ALL (RZ) geometry plots on website (all layouts, all tabs)

ghugo83:
**Tilted modules with auto placement for the Pixel :**
- Until now, width / length of a module for the flat part, could only be specified until Layer hierarchy level in cfg file.
It is now possible to specify width / length of a module for the flat part at the Ring hierarchy level in cfg file, if necessary.

- Having a zError info doesn't make sense for the flat part of the Pixel.
As a result, for the flat part of the Pixel, zOverlap info is now also displayed on the website.

**Plots :
All (RZ) geometry plots on the website displayed slightly wrong R coordinates since the beginnings of tkLayout :** 

Origin O, (X,Y,Z) is the frame of reference of the Tracker.
Let be S a sensor.
Let's call (P) the plane containing (O,Z) axis and the centre of S.
Let's call Rho vectors the projections in (XY) plane of the vectors binding origin O and the extrema points of S.
Rho vectors were directly used for calculating the radii on the plots on the website. 
They were not projected into (P), while they should be !

As a result, all radii values on the plots on the website were slightly too big !
Rho vectors are now projected into (P) !

NB : Code is operational now but debug style, I am working on improving this right now :)

### DIFF
--- a/geometries/CMS_Phase2/Baseline_tilted_200_Pixel_4_1_0.cfg
+++ b/geometries/CMS_Phase2/Baseline_tilted_200_Pixel_4_1_0.cfg
@@ -1,3 +1,3 @@
 @include-std CMS_Phase2/SimParms
-@include OuterTracker/Tilted/OT_V351_200.cfg
+@include OuterTracker/Tilted/OT_V362_200.cfg
 @include Pixel/Pixel_V4/Pixel_V4_1_0.cfg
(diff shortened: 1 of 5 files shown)


alkemyst:
Hi Gabrielle,
  thanks for the contribution. As usual I have to ask if you tested this new code against regressions.
  Also, is the display bug you described here fixed in this PR?

Thanks
S

ghugo83:
Dear Stefano,
Yes, this is tested against regressions.
This includes a fix of the display bug (include/PlotDrawer.h).
Cheers,
Gabrielle

PR #229, opened 2016-08-31
Small improvements

alkemyst:
Output improvement: table with number of modules
Usage improvement: tklayout can now be run from any directory


alkemyst:
Tested by Stefano

PR #235, opened 2016-09-12
Use of DetIds for identifying modules in Trigger towers to Modules connections (and in Modules to Trigger towers connections)

ghugo83:
- Use of the freshly implemented Detids (matching CMSSW scheme) for identifying modules in the connections : 
a) from Trigger towers to Modules.
b) from Modules to Trigger towers

- Removal of the old sebCoords identification scheme.


alkemyst:
Perfect, thanks!

PR #236, opened 2016-09-14
Added DetId info on the website

ghugo83:
DetId info appears : 
- on the geometry tabs : geometry files have an additional column with matching DetIds.
- on the info tab : summary list of all DetIds and their matching geometry info.


alkemyst:
Looks reasonable! :-)

ghugo83:
:)

PR #237, opened 2016-09-14
Removing unwanted stats from plots

alkemyst:


alkemyst:
Tested by me against regression

PR #241, opened 2016-09-29
Calculating Z and A from radiation length and nuclear interaction length by Stefano (code just copied pasted here).

ghugo83:
This is temporary, as this aspect of the code for the material will be reworked anyway.
In the meantime, it provides Z and A in an acceptable range of uncertainty.

Plus Fixed merge-ability conflict of dev_gabie into dev


alkemyst:
bhqxkz2cqaail5f

ghugo83:
Yes MaterialTable itself should be removed but shall we not take care of this when we work on the code for the material ?

alkemyst:
Thanks :-)

PR #243, opened 2016-10-05
Dev lite - xml extractors created, added full include/include-std functionality

drasal:
Several improvements implemented:
- Added tkLayout LICENCE to tkLayout-lite
- Introduced concept of XML extractors independent on experiments + filled container for FCCSW
- Added TinyXML2 functionality directly to project (XML can be exported without links to external libs)
- Added functions min/max/R/ZAllMat variables that take into account all materials inside for all hierarchical objects (layers, rods, etc.) -> used to correctly calculated object envelopes for full sim extraction
- DetectorModule class not anymore dependent on Decorator class, rather clear approach used: DetectorModule class derives from BaseModule class and thus simply incorporates Geometrical shape, called GeometricModule. Hence, each DetectorModule derived class has direct method to access its geometrical shape (GeometricModule member)
- Added full include/include-std functionality as implemented in master branch by Stefano
- Several small bugs fixed


drasal:
Once you agree with these changes, can you please merge devLite with masterLite. I consider this as a first fully working version of tkLayout-lite.?

source/Analysis/src/AnalyzerOccupancy.cc
```diff
@@ -91,29 +95,53 @@ AnalyzerOccupancy::~AnalyzerOccupancy()
...
-  fillHistogram(m_chargedMapBOnMatOnLTh,m_hisChargedFluxBOnMatOnLTh,"ChargedFluxPerPPBOnMatOnLTh","Flux of charged particles [cm^{-2}] per pp collision - B on, all material on (e-low thr.)");
-  fillHistogram(m_chargedMapBOffMatOn,  m_hisChargedFluxBOffMatOn,  "ChargedFluxPerPPBOffMatOn",  "Flux of charged particles [cm^{-2}] per pp collision - B off, all material on");
-  fillHistogram(m_chargedMapBOnMatOff,  m_hisChargedFluxBOnMatOff,  "ChargedFluxPerPPBOnMatOff",  "Flux of charged particles [cm^{-2}] per pp collision - B on, material off");
-  fillHistogram(m_chargedMapBOffMatOff, m_hisChargedFluxBOffMatOff, "ChargedFluxPerPPBOffMatOff", "Flux of charged particles [cm^{-2}] per pp collision - B off, material off");
-  fillHistogram(m_chargedMapBOffTrkOff, m_hisChargedFluxBOffTrkOff, "ChargedFluxPerPPBOffTrkOff", "Flux of charged particles [cm^{-2}] per pp collision - B off, tracker material off");
+//  fillHistogram(m_chargedMapBOnMatOnLTh,m_hisChargedFluxBOnMatOnLTh,"ChargedFluxPerPPBOnMatOnLTh","Flux of charged particles [cm^{-2}] per pp collision - B on, all material on (e-low thr.)");
```
alkemyst: I see some 'dead code' in AnalyzerOccupancy. Code that's commented out and left there. Is it possible to remove it? If not, could you please add a comment explaining if/when it should be to uncommented?

source/Analysis/src/AnalyzerOccupancy.cc
```diff
@@ -150,39 +178,39 @@ bool AnalyzerOccupancy::visualize(RootWSite& webSite)
...
-    anImagePhotonsBOffTrkOff.setComment("RZ view of photons flux [cm^-2] in a tracker - B off,  tracker material off (calorimeter on)");
-  }
-  if (drawHistogram(canvasPhotonsBOffMatOn, m_hisPhotonsFluxBOffMatOn, m_photonsMapBOffMatOn, "PhotonsCanvasBOffMatOn", "RZ view of photonsflux")) {
-    canvasPhotonsBOffMatOn.SetLogz();
-    m_hisPhotonsFluxBOffMatOn->SetMinimum(c_fluxMin);
-    m_hisPhotonsFluxBOffMatOn->SetMaximum(c_fluxMax);
-    RootWImage& anImagePhotonsBOffMatOn = plotsPhotonsContent.addImage(canvasPhotonsBOffMatOn);
-    anImagePhotonsBOffMatOn.setComment("RZ view of photons flux [cm^-2] in a tracker - B off,  all material on");
-  }
+//  TCanvas canvasPhotonsBOffMatOn;
```
alkemyst: [more dead code]

source/Analysis/src/AnalyzerOccupancy.cc
```diff
@@ -196,39 +224,39 @@ bool AnalyzerOccupancy::visualize(RootWSite& webSite)
...
-    anImageChargedBOffTrkOff.setComment("RZ view of charged particles flux [cm^-2] in a tracker - B off,  tracker material off");
-  }
-  if (drawHistogram(canvasChargedBOffMatOn, m_hisChargedFluxBOffMatOn, m_chargedMapBOffMatOn, "ChargedCanvasBOffMatOn", "RZ view of charged particles flux")) {
-    canvasChargedBOffMatOn.SetLogz();
-    m_hisChargedFluxBOffMatOn->SetMinimum(c_fluxMin);
-    m_hisChargedFluxBOffMatOn->SetMaximum(c_fluxMax);
-    RootWImage& anImageChargedBOffMatOn = plotsChargedContent.addImage(canvasChargedBOffMatOn);
-    anImageChargedBOffMatOn.setComment("RZ view of charged particles flux [cm^-2] in a tracker - B off,  all material on");
-  }
+//  TCanvas canvasChargedBOffMatOn ;
```
alkemyst: [more dead code]

source/Analysis/src/AnalyzerOccupancy.cc
```diff
@@ -237,203 +265,203 @@ bool AnalyzerOccupancy::visualize(RootWSite& webSite)
...
-    m_hisChargedRatioLTh  = (TH2D*)(m_hisChargedFluxBOnMatOnLTh->Clone("ChargedFluxPerPPRatioLTh"));
-    m_hisChargedRatioLTh->Divide(m_hisChargedFluxBOnMatOn);
-    m_hisChargedRatioLTh->SetTitle("Ratio of charged particles flux - (Cut_{e prod}=10keV, Cut_{#gamma prod}=1keV)/(Cut_{e prod}=1MeV, Cut_{#gamma prod}=100keV)");
-
-    if (drawHistogram(canvasChargedRatioLTh, m_hisChargedRatioLTh, m_chargedMapBOnMatOn, "ChargedRatioLTh", "RZ ratio of charged particle fluxes")) {
-      RootWImage& anImageChargedRatioLTh = plotsChargedRatioContent->addImage(canvasChargedRatioLTh);
-      anImageChargedRatioLTh.setComment("RZ ratio of charged particles fluxes - (Cut_{e prod}=10keV, Cut_{#gamma prod}=1keV)/(Cut_{e prod}=1MeV, Cut_{#gamma prod}=100keV)");
-    }
-  }
+////  // Ratios
```
alkemyst: [more dead code]

source/Geometry/src/Disk.cc
```diff
@@ -9,8 +9,8 @@ Disk::Disk(int id, double zOffset, double zECCentre, double zECHalfLength, const
  zError(        "zError"     , parsedAndChecked()),
  m_innerRadius( "innerRadius", parsedAndChecked()),
  m_outerRadius( "outerRadius", parsedAndChecked()),
- m_bigDelta(    "bigDelta"   , parsedAndChecked()),
- m_rOverlap(    "rOverlap"   , parsedOnly(), 1.),
+ bigDelta(      "bigDelta"   , parsedAndChecked()),
```
alkemyst: I see that smallDelta and rOverlap were renamed to their original name from m_smallDelta and m_rOverlap. I think that it was more convenient to keep those member variables names the same as their string counterparts, but since you decided to rename all of them, why did you drop the m_ for these two?

source/Geometry/src/GeometricModule.cc
```diff
@@ -1,172 +1,210 @@
...
-  }
-
-}
-
-
-
-
-
-
+//double GeometricModule::trackCross(const XYZVector& PL, // Base line point
```
alkemyst: Also here there is a large chunk of 'dead code'. Where did the corresponding functionalities go? Can we remove these lines?

source/Geometry/src/Layer.cc
```diff
@@ -24,12 +26,12 @@ Layer::Layer(int id, int barrelNumLayers, bool sameRods, bool barrelMinRFixed, b
  sameParityRods (       "sameParityRods"     , parsedAndChecked(), true),
  layerRotation  (       "layerRotation"      , parsedOnly()      , 0.),
  tiltedLayerSpecFile(   "tiltedLayerSpecFile", parsedOnly()),
- m_smallDelta   (       "smallDelta"         , parsedAndChecked()),
+ smallDelta     (       "smallDelta"         , parsedAndChecked()),
  m_smallParity  (       "smallParity"        , parsedAndChecked(),-1),
```
alkemyst: Also here: I personally prefer smallParity to m_smallParity, but then if we want to be consistent, why not m_smallDelta?

drasal:
Stefano,
I still need to work out your comments. I will do it in the next pull request then! (after Wednesday's FCC-hh meeting)

alkemyst:
Hi Zbymek,
   when would you like to meet?

'S

-- 
Stefano Mersi
EP-CMX-DA
CERN - European Organization for Nuclear Research
   1211 Geneva 23 - Switzerland
   Mobile +41 75 411 5816
   Office +41 22 767 4749
   Fax    +41 22 766 8631
   Mail stefano.mersi@cern.ch

drasal:
Hi Stefano,

I would be grateful for anytime today or tomorrow... Please, let me know if you manage to find a time slot...

 Thanks a lot,
 Zbynek


________________________________

source/Geometry/src/Disk.cc
drasal: The idea behind this was to keep the private variables as e.g. m_rOverlap(), but the public as rOverlap() -> simply applying the coding conventions. The configuration doesn't care about any coding conventions, hence the string counterpart is without m_ . Just let me know, what the procedure should be and I will follow that in the whole code... It's matter of taste... If we want to keep them without m_ I will make all public.

source/Geometry/src/GeometricModule.cc
drasal: This code was originally there and no code of the tklayout uses that, hence I commented that out to avoid any clashes in the future. I didn't want to put the code directly away ... If you want, I can leave uncommented, but somebody in the future may start thinking, why it's there...

source/Geometry/src/Layer.cc
drasal: The same comment as for the m_bigDelta() applies here, we simply need to agree on the "coding" conventions for this special variables, which also work as functions ...

drasal:
We noticed that material envelope is necessary also for whole barrels/endcaps, hence another "bug" fix for the Extractor & implementation of min/maxZ/RAllMat() functions. This time for barrel/endcaps...

drasal:
We noticed that material envelope is necessary also for whole barrels/endcaps, hence another "bug" fix for the Extractor & implementation of min/maxZ/RAllMat() functions. This time for barrel/endcaps...

drasal:
The occupancy analyzer was updated -> I've removed the commented lines and made the analyzer more general for any experiment studies (the previous version was too FCC specific). In addition we have added small adjustments to FCCSWExtractor to be more compatible with full simulations (DD4Hep Sensitive class segmentation & cellID numbering scheme).

ghugo83:
Hi Zbynek,
About the point on global extrema for barrel / endcaps containers.
You might want to use sth else for defining the outermost (entire Barrel or entire Endcaps) envelopes (or maybe not, I am just mentioning why this was needed for CMS :) ).
The thing is, as you know, there are a few services going out of the Endcaps in Z, but located at a small [r; r+dr] region only for example.
If you take the 'global' extrema, you get a volume much bigger than needed.
In CMS, this increases considerably the simulation time. I do not know about FCC software though ?

This is why a Polycone was needed (basically like a plastic bag wrapping an object with low air pressure inside the bag ;p).
But again, I do not know whether this matters for FCC,
Cheers

drasal:
Hi Gabi,

thanks for this point! From my experience I don't see a problem with the bigger volume, unless one doesn't force Geant4 to use artificially small step. Geant4 is clever enough to adjust the step (& thus the speed) according to the material traversed. So, if the envelope filled with gas is empty, it shouldn't be a problem, which is exactly what we see in full sim. On the other hand, I will keep you comment in my head. We currently don't output the services, so we need to test this functionality in the future ... I assume that the general function providing the global max/min can stay in the SW, we will just need it for the Extractor? Or is it colliding with anything?

 Cheers, Zbynek

________________________________

alkemyst:
Hi Zbynek,
  are you happy with this PR as it is now? Can I merge it?

drasal:
Hi Stefano,

the comment, which Gabi had is of practical impact rather than of bad coding inside the SW. So, I have to keep it in my head once improving the FCCSW extractor, but it doesn't break any SW functionality ...

As regards the property variables and their names. Following the coding conventions, I named these variables in a way, which may look weird at first sight, but simply: m_name means private member property variable, name without m_ means public property variable. If you feel that it's too much complex compared to what we had before, don't hesitate to let me know and I will rename it. If the program is then more easy to be followed, it makes sense the make exception for property variables. Please, let me know, your preferences. I will change it then. If you're fine with that, I feel that this PR is OK.

PR #244, opened 2016-10-31
Creating version 3.6.3

alkemyst:


alkemyst:
En passant, also adding a subtotal in the module count that was missing

alkemyst:
Tested by Stefano

PR #252, opened 2016-11-18
Modified definition of zOverlap in tilted rings placement + added possibility to reverse the process of tilted modules placement (ringInnerZ or ringOuterZ in input, zOverlap as output)

ghugo83:
- Definition of zOverlap in tilted rings placement : 
Until now, when ring i+1 is placed with respect to ring i : the mean of the zOverlaps of both adjacent rings is used. 
Now, zOverlap is redefined in the code, so that when the ring i+1 is placed with respect to ring i : the zOverlap of ring i+1 only is used.

- Added possibility to reverse the process of tilted modules placement :
Until now : 
zOverlap set as input.
both ringInnerZ and ringOuterZ as output.
What was added : 
ringInnerZ or ringOuterZ set as input (not both).
zOverlap as output.

- zOverlap parameter, used for tilted rings placement, is renamed to ringZOverlap.
This is to avoid that zOverlap from a Layer is used accidently for tilted rings.
This also allows cfg files using the defintion of zOverlap used until now, to be recognized.

- The formula of phiOverlap is unchanged. 
Though, since phiOverlap depends on zOverlap value, new phiOverlap values are obtained.
They correspond to a new definition : phiOverlap is the overlap of modules in phi, seen from the origin, at edge of the module - ringZOverlap (previously, a much less accurate zOverlap was withdrawn).

- All TBPS cfg files are modified accordingly.
Same coordinates should be obtained for all layouts (already checked, but please cross-check).


alkemyst:
The new configuration files using ringInnerz in place of zOverlap were tuned to give the same module positions for the following TBPS layouts
TBPS_V360_200.cfg
TBPS_V361_200.cfg
TBPS_V362_200.cfg
TBPS_V363_200.cfg
TBPS_V364_200.cfg
TBPS_V462_200.cfg

This was tested and the module positions are the same within a few nanometers.

alkemyst:
Core reviewed by Stefano and Gabrielle

ghugo83:
Additional tests : 
- zOverlap < 0
- zOverlap > (length / 2.)    with length = length of the module
- tilted pixel
All show expected behaviour of the newly defined ringZOverlap.

PR #254, opened 2016-11-22
Corrected ommission in straight rod module placement, in particular case where zEnd < zError and the build is upwards 

ghugo83:
This is a subtle diff in the straight rod module placement, when building ring (i+1) with respect to ring (i) considering the zError constraint.
In the particular case where zEnd of the module in ring (i) < zError, and the build is upwards, one needs to consider : 
* -dsdistance/2. for ring (i)
* +dsdistance/2. for ring (i+1)
Indeed, this corresponds to the most stringent situation.

For all other cases, one needs to consider (and this was already the case) : 
* +dsdistance/2. for ring (i)
* -dsdistance/2. for ring (i+1)

The effect of this, for example on the flat part of TBPS, is (example with OT 364) : 
- Layer 1 : ~ 120 um shorter.
- Layer 2 : ~ 26 um shorter.
- Layer 3 : ~ 13 um shorter.

This is negligible, but incorporated at this point in time for safety reasons.

### DIFF
--- a/config/stdinclude/CMS_Phase2/Pixel/Resolutions/README.txt
+++ b/config/stdinclude/CMS_Phase2/Pixel/Resolutions/README.txt
@@ -1,5 +1,5 @@
-This is to simulate resolution on pixel modules local coordinates X and Y.
-resolutionLocalX or Y are either nominal values, either values calculated from parameterized models.
+This is to simulate resolution on pixel modules local coordinates X and Y with tkLayout.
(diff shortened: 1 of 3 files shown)


ghugo83:
The influence of this fix on TB2S is understood : this introduces a diff of up to 0.2 mm in TB2S when smallParity = -1 (smallParity = -1 not present in any of our cases of study though).
It would also have introduced a similar order of diff in the flat TBPS barrels as well.

PR #256, opened 2016-11-24
Reverted #255

ghugo83:
Reverts tkLayout/tkLayout#255
Code compiles, but segfaults when running.
Shall we revert this PR on the Makefile, the time we do the corresponding fixes on the Makefile on it ?


alkemyst:
Yes, very good idea.

PR #258, opened 2016-11-24
IMPORTANT: fixing pixel capacitor config file

alkemyst:


alkemyst:
Checked by Stella+Gabrielle, it works :-)

PR #260, opened 2016-12-01
Merging gabrielle's "material" branch into dev_gabie

alkemyst:


include/Analyzer.h
```diff
@@ -155,7 +166,7 @@ namespace insur {
                                const std::vector<double>& momenta,
                                const std::vector<double>& triggerMomenta,
                                const std::vector<double>& thresholdProbabilities,
-			       bool isPixel,
+			       bool isOuterPixel,
```
alkemyst: @ghugo83 : are you sure about this line?

ghugo83: Hmm you are right, this was probably accidentally written when looking for word 'Outer'.
This has absolutely no effect though, since in this context only the type matters.
I just fixed this typo though ;)

ghugo83: Thanks for noticing it :)

ghugo83:
As a general note, this is a very quick-and-dirty style code (sorry ;p).
Results for the material budget within tracking volumes have been deeply cross-checked and compared with full volume + CMSSW results.
Though, the code itself will **really** need some cleaning !!!!


This PR includes : 
- Updated Beam Pipe Material Budget estimates by tkLayout. 
Radiation length : tkLayout MB was ~1.5% higher.
Interaction length : tkLayout MB was ~7% lower.
There is now less than 0.5 % discrepancy between tkLayout and CMSSW  as regards Beam Pipe MB.
NB : CMSSW Beam Pipe MB includes air material.

- Assigned categories to hits. That is to say, does the hit belong to : 
BeamPipe ?
Pixel Interstice ? (=between BeamPipe and PixelTrackingVolume)
Pixel TrackingVolume ? (=between first pixel active hit and last pixel active hit)
Interstice ? (=between PixelTrackingVolume and OuterTrackingVolume)
OT TrackingVolume ? (=between first OT active hit and last OT active hit)

- Estimated Material budget within Tracking Volume. 
On the website, results can be shown per components type. Totals are also shown.
Results are presented in the different tabs : 
* Info tab : full tracking volume.
(Beam pipe + Pixel Interstice + Pixel Tracking Volume + Interstice + OT Tracking Volume).
* Pixel tab : Pixel Tracking Volume.
* Outer Tracker tab : OT Tracking Volume.

- Detailed services (routed materials) in an automatic way, which is very time-earning.

- All Material Budget plots (full volume + tracking volume) can now have a number of bins simply set in Vizard.h (materialNBins=300 by default).

- General presentation improvements of plots for materials on website.

- Fixed several bugs that appear in conversion stations.

ghugo83:
Validation : 
Did similar hacks in code for the material in CMMSW to get MB per tracking volume.

Results of validation: 
- MB within full volume : less than ~2% discrepancy between CMMSW and tkLayout.
- MB within tracking volume : tkLayout MB is ~5% higher. This is explained by the modules materials, which are not placed exactly at the same location in CMSSW and in tkLayout.

PR #274, opened 2016-12-14
Latest devs in the XML export in sync with CMSSW + Added layout FlatTracker4026

ghugo83:
After the merge of these commits, the code in tkLayout/dev_gabie will be in the exact state of what was used for the XML export of FlatTracker4026 and TiltedTracker4026 to CMSSW.

Updates in the XML export : 
Worked on the materials XML export.
- 'Elementary' materials : 
They are now in tracker.xml only (they were duplicated in pixel.xml).
They are prefixed by "tkLayout_" for additional safety reasons.

- Composite materials (that is to say, all the other materials !) : 
They used to be a lot of composites with different names but same actual materials compositions. Redundancies have been deleted.
Design has been chosen towards the safest solution : scan the composite materials once they are created, and print the duplicated materials only once in the XMLs.
First glance solution that module materials could be categorized (category 'module type + dsDistance' for example, or similar), turned out to be impossible from the materials point of view, and is also much less safe.

Other changes in the XML export : 
- updated one constant in trackerProdCuts.xml, after question to Vladimir.
- deleted a few obsolete references to Phase 1 Pixel.

Added FlatTracker4026.
The soup recipe is the following : 
- TBPS from Flat/OT_200.cfg
- TB2S, TEDD from OT_Tilted_365
- Pixel is Pixel4026

### DIFF
--- a/src/Extractor.cc
+++ b/src/Extractor.cc
@@ -1898,12 +1898,10 @@ namespace insur {
       if ((iter->getZOffset() + iter->getZLength()) > 0) c.push_back(createComposite(matname.str(), compositeDensity(*iter), *iter));
 #else
       matname << xml_base_serfcomp << "R" << (int)(iter->getInnerRadius()) << "Z" << (int)(fabs(iter->getZOffset() + iter->getZLength() / 2.0));
-
       shapename << xml_base_serf << "R" << (int)(iter->getInnerRadius()) << "Z" << (int)(fabs(iter->getZOffset() + iter->getZLength() / 2.0));
       if ((iter->getZOffset() + iter->getZLength()) > 0 ) {
         if ( iter->getLocalMasses().size() ) {
           c.push_back(createComposite(matname.str(), compositeDensity(*iter), *iter));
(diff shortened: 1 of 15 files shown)


src/Extractor.cc
```diff
@@ -3091,23 +2992,22 @@ namespace insur {
   }
 
   void ModuleComplex::addMaterialInfo(std::vector<Composite>& vec) {
-    std::vector<Volume*>::const_iterator vit;
-    for ( vit = volumes.begin(); vit != volumes.end(); vit++ ) {
-      if ( !((*vit)->getDensity()>0.) ) continue; 
+    for (const auto& vit : volumes) {
+      if (!(vit->getDensity() > 0.)) continue;
```
alkemyst: Maybe we can throw an INFO or WARN here?

src/XMLWriter.cc
```diff
@@ -574,27 +574,53 @@ namespace insur {
...
+      for (const auto& elem : elements) {
+	stream << xml_material_fraction_open << elem.second << xml_material_fraction_inter;
+	stream << xml_fileident << ":" << xml_tkLayout_material << elem.first << xml_material_fraction_close;
+      }
+      stream << xml_composite_material_close;
+
+      // Special case where a composite with same name already exists
+      if (mapCompoToPrintedCompo_.find(nspaceName) != mapCompoToPrintedCompo_.end()) {
+	//throw PathfulException("Found several composite materials with same name " + nspaceName);
+	std::cout << "VERY IMPORTANT : VOLUME " << nspaceName << " IS DUPLICATED !!!!!!!!!!" << std::endl;
```
alkemyst: Can we change this to a logSOMETHING ?

PR #280, opened 2017-01-13
Extended irradiation power maps to Pixel + Added FLUKA 1-MeV-neutron fluence maps generated in Phase 2 conditions + Fixed bugs in irradiation power maps code

ghugo83:
- Extended irradiation power maps to Pixel.

- Created 1-MeV-neutron fluence maps (Phase 2) with FLUKA and added them to tkLayout.
They replaces the former FLUKA maps (Run 2 ).
This is of importance, since the preshower has been replaced by a new shielding of Ecal, resulting in a significantly lower fluence at the most forward disks of TEDD.
Hence, the sensor irradiation power, on tkLayout irradiation maps, is updated from ~0.8 W to ~0.55 W for a few TEDD rings for example.
See resulting irradiation maps on tkLayout website.

- Fixed bug in 2D linear interpolation for calculating fluence from FLUKA maps.

- Fixed bug due to incorrect choice of observation points for the modules sensors (error lower than 5% though).

- Possibility to set module-dependent parameters (such as operating temperature for example) by a new @include-std CMS_Phase2/Pixel/moduleOperatingParms
(These parameters used to be in SimParms).

NB :  The pixel sensor thickness is not taken into account in the enveloppePoly, which has no relevant effect here but will be fixed soon.


alkemyst:
c1
Variation of evaluated irradiation on Outer Tracker sensors (1=no change)

src/AnalyzerVisitors/IrradiationPower.cpp
```diff
@@ -1 +1,69 @@
...
+
+void IrradiationPowerVisitor::visit(DetectorModule& m) {
+  operatingTemp    = m.operatingTemp() + insur::celcius_to_kelvin;
+  chargeDepletionVoltage = m.chargeDepletionVoltage();
+  double volume = 0.;
+  std::vector<double> irradiationValues;
+  
+  for (const auto& s : m.sensors()) {
+    // Calculate total volume occupied by sensors
+    volume += s.sensorThickness() * m.area() / 1000.0; // volume in cm^3
```
alkemyst: Please use units here: instead of /1000.0 it should be (probably) (/Unit::mm*Unit::cm)^3

ghugo83:
- few renamings as discussed.
- if no biasVoltage or operatingTemp is specified in cfg file, an error is thrown indeed.
- worked on polygons operations, so that more points can be directly accessible from class Sensor.
- notably, for calculating the irradiation power, the points, located at the middle between two sensor parallelepiped vertices, need to be taken into account.
- this can also allows to have correct minR/maxR/minZ/maxZ for the Pixel (sensorThickness was taken into account correctly for calculating OT extrema but not for Pixel). The diff is obviously negligible, but in practice for calculating envelopes / avoiding overlaps, this is useful.

ghugo83:
Done ;)

PR #283, opened 2017-01-24
Added a split between rods located at +bigDelta or -bigDelta in sensors irradiation power results

ghugo83:
Until now, irradiation results for a given barrel layer, was not taking into account whether the rods are located at +bigDelta or -bigDelta.
It was randomly either of the 2 rods which was used for the results in tables on the website.

Even if the uncertainty on FLUKA data, which is used for the results, is 10% anyway, we have here a too strong simplification.
The new, more detailed, results (split tables by inner radius / outer radius rods) show :  
In the worse case (central ring of the lowest layer) : 
TB2S : -6%
TBPS : ~ +17%
Pixel barrel :  ~+19% 
(In the worse case, the diff is big enough to be worth being taken into consideration).

In the best case : 
TB2S : -1%
TBPS : ~ +2%
Pixel barrel :  ~+2% 
(In the best case, diff has no interest).

### DIFF
--- a/include/AnalyzerVisitors/IrradiationPower.h
+++ b/include/AnalyzerVisitors/IrradiationPower.h
@@ -14,6 +14,8 @@
 #include "SummaryTable.h"
 
 typedef std::tuple<bool, bool, std::string, int, int> ModuleRef;
+// Used to identify a module.
+// bool isBarrel, bool isOuterRadiusRod, std::string barrel/endcap name, int layer/disk number, int ring number
 
 class IrradiationPowerVisitor : public GeometryVisitor {
   double timeIntegratedLumi_;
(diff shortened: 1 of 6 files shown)


alkemyst:
This is really neat. Thanks +1 :-)
I am not sure from looking at the code: do tilted rings also get split in inner and outer?

ghugo83:
Yes, basically, a tilted rod contains a flat part and a tilted part, as follows.

The 'inner radius' rod contains (at a given Phi) : 
-flat part : the modules located at -bigDelta.
-tilted part : the inner modules of the tilted rings.

The 'outer radius' rod contains (at a given Phi) : 
-flat part : the modules located at +bigDelta.
-tilted part : the outer modules of the tilted rings.

PR #293, opened 2017-02-24
Fixing an old bug

alkemyst:
trigger tuning analysis would crash if non-pt modules are in the Outer Tracker


alkemyst:
tested by Stefano: fixes the issue on SLC6

PR #295, opened 2017-03-02
Fixing a time-wasting bug

alkemyst:
the "dead code" errorSummary was called even when taggedErrorSummary should have been
  TODO: remove *all* code under NO_TAGGED_TRACKING


alkemyst:
Tested by Stefano

PR #302, opened 2017-03-10
Dev detailed endcap plots

alkemyst:


alkemyst:
Sorry: this PR contains too much stuff. I am going to create a new one with just the detailed endcap plots

PR #303, opened 2017-03-10
Detailed plots of the different endcaps

alkemyst:
This gives a warning I do not  understand during compilation of Vizard.o:
What I do not understand is why is anEndcap const...

Building target Vizard.o...
colorgcc -std=c++11  -g -fpermissive -lstdc++ -fmax-errors=2 -Iinclude/ `root-config --cflags` -c -o lib/Vizard.o src/Vizard.cc
src/Vizard.cc: In member function ‘void insur::Vizard::createSummaryCanvasNicer(Tracker&, TCanvas*&, TCanvas*&, TCanvas*&, std::vector<TCanvas*>&)’:
src/Vizard.cc:6066:47: warning: passing ‘const Endcap’ as ‘this’ argument discards qualifiers [-fpermissive]
       set<Module*> modules = anEndcap.modules();
                                               ^
In file included from include/Tracker.h:17:0,
                 from include/Vizard.h:44,
                 from src/Vizard.cc:6:
include/Endcap.h:78:16: note:   in call to ‘std::set<DetectorModule*> Endcap::modules()’
   set<Module*> modules();

### DIFF
--- a/include/Disk.h
+++ b/include/Disk.h
@@ -125,6 +125,7 @@ public:
   const MaterialObject& materialObject() const;
   ConversionStation* flangeConversionStation() const;
   const std::vector<ConversionStation*>& secondConversionStations() const;
-};
+  std::vector< std::set<const Module*> > getModuleSurfaces() const;
+ };
 
 #endif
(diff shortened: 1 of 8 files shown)


alkemyst:
Following Gabrielle's suggestion, I revised this code, implementing a visitor inside PlotDrawer

PR #304, opened 2017-03-13
Quick code to fix the issues in rings placement in TEDD

ghugo83:
Modified the rings placement in TEDD, which had issues.

For a given TEDD : 

1) Vector of dsDistances (for a disk) not taken into account properly.
 For a given disk, instead of the correct vec of dsDstances, the vec of maxDsDisances was used. 

2) Vector of smallDeltas (for a disk) not taken into account properly.
For example, for OT567 , in TEDD_2 :
- Disks 1 and 2 rings radii were calculated, with the constraints that Disk 1 Ring 12 smallDelta = 7.45 (fine) and Disk 3 Ring 12 smallDelta = 7.45 (not fine).
- Disks 3 rings radii were calculated, with the constraints that Disk 1 Ring 12 smallDelta = 8.55 (not fine) and Disk 3 Ring 12 smallDelta = 8.55 (fine).
**Consequences : Disk within TEDD_2 were differnet (different radi and even different number of modules).**

NB : Very quick code, style to be improve, but results fully tested.

### DIFF
--- a/include/Disk.h
+++ b/include/Disk.h
@@ -44,9 +44,10 @@ private:
   PropertyNode<int> ringNode;
   PropertyNodeUnique<std::string> stationsNode;
 
-  inline double getDsDistance(const vector<double>& buildDsDistances, int rindex) const;
-  void buildTopDown(const vector<double>& buildDsDistances);
-  void buildBottomUp(const vector<double>& buildDsDistances);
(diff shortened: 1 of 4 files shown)


alkemyst:
Thanks a lot

PR #315, opened 2017-03-16
New module count table

alkemyst:
split by module type (incl. sensor spacing) and subdetector

Using a new Report paradigm, with analyze on tracker(s) and visualize separately


alkemyst:
Taking some comments from Gabrielle into account

PR #321, opened 2017-03-21
Dev gabie - implemented new tracking functionalities

drasal:
1) Implemented new Tracking functionality from tkLayout-lite tkLayout.

 * Practically, new classes: TrackNew and HitNew implemented.
 * For compatibility reasons the original hit class was not removed, the tracking functionality commented out in this class to avoid any confusion.
 * SimParms class changed to singleton pattern. Magnetic field defined in SimParms config file now.
 * Property class updated -> readonly property object need to have functionality to assign them correct unit.
 * Track & Hit classes require correct units, therefore, mainConfigHandler assign correct units to read-in momenta.
 * All other changes not related to tracking connected with change of SimParms to singleton pattern.

2) Implemented pattern recognition analysis & visualization from tkLayout-lite 
* Currently only inside-out approach supported, outside-in will follow soon.
* Added fluence per 1pp collision map.


ghugo83:
Hi Zbynek,
Thanks a lot !
In my local it doesn't compile anymore though. You didn't add TrackNew and HitNew to the Makefile ?

drasal:
Hi Gabi,

unfortunately I'm not a  Makefile guy, I've added everything to CMake. Can you or Stefano, please change the Makefile?

ghugo83:
Sure, I ll do that and add the commit to ur branch

drasal:
Thanks a lot!

ghugo83:
It works with the fix, I did the commit and PRed it to your branch

drasal:
Added several minor improvements regarding TrackNew class:
* IP spot defined with units in SimParms. rError renamed to rphiErrorCollider to have the same name as its equivalent in Z.
* Removed passage of track parameters in Vizard & Analyzer methods, if track passed as parameter too (The track already provides such parameters.)

ghugo83:
Hi Zbynek,
Greeeat :) Just one little thing : why Hit and NewHit classes in parallel ? 
Do I understand correctly that the NewHit is for tracking functionalities, and that part of the code in Hit will need to be moved to NewHit at some point ?
Cheers

drasal:
Hi Gabi,

a good point. Too many Analyzer parts use standard Track with standard Hits for different purposes (material budget etc.), so complete remove of hit class would need lots of changes. Hence, I've decided to go slowly step-by-step and keep TrackNew & HitNew classes for tracking & pattern recognition. We should update other analyzer & vizard parts as well and then, we may comfortably remove the hit.cc class completely.

alkemyst:
Hi Zbynek,
   could you please summarize what is the main difference between the 
two classes?
Thanks
S

drasal:
So, the hit.cc class was separated to standalone Track class & Hit class and new functionality related to track propagation has been implemented. What is different from the original hit class except the functionality is the way how it fulfills the new C++ standards. No raw pointers are allowed, no return of raw pointer is allowed, only references or const pointers. Internally, the actual implementation uses unique smart pointer techniques... There is a straight-forward way to use the new Track and Hit class in the other pieces of the code, but quite some stuff would be affected then. This may need a separate testing for each analysis part of the code. As my eye is not experienced with CMS Ph2 setup, I wanted to avoid this. So that's basically the reason, why I used these classes, where really needed, so in resolution estimation & pattern recognition.

We may continue in adapting the code slowly by using these classes...

drasal:
Does this make sense?

drasal:
I would suggest to run the same configuration files with older code and the new code and check the resolution plots. Myself, I don't see any deviation from the old results in resolution plots....

alkemyst:
Hi Zbynek,
  indeed there is a 10% difference in the pT resolution for low pT momenta.
  You can clearly check this out here: I froze the dev_gabie results before the update and after and you can compare the results here:

http://cms-tklayout.web.cern.ch/cms-tklayout/layouts/repository-git-dev_gabie_old/OT365_200_IT4024/errorstracker.html
http://cms-tklayout.web.cern.ch/cms-tklayout/layouts/repository-git-dev_gabie/OT365_200_IT4024/errorstracker.html

  Is this understood?

Thanks
Stefano

alkemyst:
Another comment (sorry for flooding)
The trigger performance estimation was broken by this PR:

http://cms-tklayout.web.cern.ch/cms-tklayout/layouts/repository-git-dev_gabie_old/OT365_200_IT4024/triggerPerf.html
http://cms-tklayout.web.cern.ch/cms-tklayout/layouts/repository-git-dev_gabie/OT365_200_IT4024/triggerPerf.html

image
image

I guess it's a trivial thing, but could you please have a look?

drasal:
As for the difference in p, I have an idea, where it comes from. Let me check... As for the trigger, I'm afraid, I may need your help...

drasal:
Hi Stefano,

the difference by 10% for the low pT is coming from the following lines in TrackNew.cc class:
* Lines 1078-1080:
`   double A = 0;
     if (SimParms::getInstance().isMagFieldConst()) A = m_hits.at(i)->getRPos()/2./getRadius(m_hits.at(i)->getZPos());     // r_i/2R
     double corrFactor = 1 + A*A*cos(m_theta)*cos(m_theta)/(1-A*A);`
* Lines 1309-1311:
`    double A = 0;
     if (SimParms::getInstance().isMagFieldConst()) A = m_hits.at(i)->getRPos()/2./getRadius(m_hits.at(i)->getZPos());  // r_i/2R
     double corrFactor = pow( cos(m_theta)*cos(m_theta)/sin(m_theta)/sqrt(1-A*A) + sin(m_theta) ,2); // Without correction it would be 1/sin(theta)^2`

This correction takes into account a propagation of MS error on virtual barrel plane, on which all measurements are evaluated for consistency (global chi2 fit applied) ->in limit R->inf. propagation along line is used, otherwise a very small correction factor coming from the circular shape of particle track is necessary (THAT's the NEW STUFF)

alkemyst:
Hi Zbynek,
  for the trigger part I prepared PR (#328) which fixes the problem. In the configuration file trigger momenta are meant to be in units of MeV :-)

alkemyst:
Hi all,
  please trike my previous comment on MeV/GeV...

drasal:
It's in the ptErrorAdapter code, uhhh that's pretty ... Full of hard-coded numbers. Let's see if I manage to do it properly!

alkemyst:
Hold on a second first. I think I know how to best proceed.
I will first make another PR reverting your attempt to introduce Units in mainConfigHandler. That fixed everyhting.

I would then try to slowly kill mainConfigHandler altogether, as we can move everything into SimParms. When we'll do so we will also fix the places where it is used.

As it stands now it it even difficult to debug, while if we use SimParms it would be straightforward.

alkemyst:
Ok, I fixed this in PR #330 (removing Units.h from mainConfigHandler)
I will then proceed to kill it :-)

drasal:
Hi Stefano, 

you don't have to do it. I'm almost there... I'd keep the units in the game, as it makes the code more readable.

PR #325, opened 2017-03-22
Removing a redundant code in makefile to compile TrackNew and HitNew

alkemyst:


ghugo83:
Haha how did that happen ?
At the time of this morning when I did commit b1a5d3e615cbf661b323fa447704941f795c3e9b they were not there at all though, and i added TrackNew and HitNew only once each

alkemyst:
Because I had a private branch in which I was already merging Zbynek's new code adn used that one to push my contributions

PR #326, opened 2017-03-22
Dev gabie - added outside-in pattern recognition

drasal:
Several changes:
* Fix to ptError calculations (trigger output) -> correctly implemented default units. Necessary to make compatible the previous code with trigger output!
* Updated Pattern recognition module with new functionality - outside-in approach


drasal:
I've shared the updated results with you by CERNbox. Feel free to have a look...

drasal:
I've reverted your change in mainConfigHandler -> units restored.

alkemyst:
Tested and working fine.

PR #327, opened 2017-03-22
Cleaned and improved style in recent geometry code for TEDD

ghugo83:


alkemyst:
Hi Gabrielle,
  is there any difference in the output for this PR?

ghugo83:
No, I checked the geometry tab and exact same coordinates (run on cfg files where coordinates are not set in them obviously ;))

PR #330, opened 2017-03-23
Reverting Uits in mainConfigHandler, as it breaks triggerMomenta, and…

alkemyst:
… also it needs to be killed anyway :-0


drasal:
Stefano, let me know, if you find any other discrepancies... I'm willing to spend today to fix all the stuff, such as we don't encounter any issues in the future, when I may be busy with other stuff...

PR #344, opened 2017-03-28
Adding the position of hottest modules

alkemyst:


src/AnalyzerVisitors/IrradiationPower.cc
```diff
@@ -67,31 +67,44 @@ void IrradiationPowerVisitor::visit(DetectorModule& m) {
...
+  sensorsIrradiationPerType.setCell(0, 3, "95% module irrad (avg) [Hb]");
+  sensorsIrradiationPerType.setCell(0, 4, "z_max [mm]");
+  sensorsIrradiationPerType.setCell(0, 5, "r_max [mm]");
   int iRow=0;
   for (auto& it : mapTypeToIrradiation_ ) {
     iRow++;
     const std::string& typeName = it.first;
-    std::vector<double>& irrads = it.second;
-    std::sort(irrads.begin(), irrads.end());
+    std::vector<const DetectorModule*>& irrads = it.second;
```
ghugo83: irrads : 
Why not using a std::set here ?
Also, why not simply having double instead of DetectorModule ?

src/AnalyzerVisitors/IrradiationPower.cc
```diff
@@ -67,31 +67,44 @@ void IrradiationPowerVisitor::visit(DetectorModule& m) {
...
+		return a->sensorsIrradiationMean() < b->sensorsIrradiationMean();
+	      });
     int nModules = irrads.size();
     std::ostringstream irrad_Max("");
     std::ostringstream irrad_95perc("");
-    irrad_Max    << std::dec << std::scientific << std::setprecision(2) << irrads.at(nModules-1);
-    irrad_95perc << std::dec << std::scientific << std::setprecision(2) << irrads.at(nModules*95/100-1);
+    std::ostringstream max_z("");
+    std::ostringstream max_r("");
+    auto& hottestModule =  irrads.at(nModules-1);
```
ghugo83: In case irrads is empty (could happen that we have no module because of some reason / issue), we will have trouble here.
I would add a check that irrads size is not 0.

ghugo83: Use end() ?

src/AnalyzerVisitors/IrradiationPower.cc
```diff
@@ -67,31 +67,44 @@ void IrradiationPowerVisitor::visit(DetectorModule& m) {
...
+    auto& hottest95Module = irrads.at(nModules*95/100-1);
```
ghugo83: This takes the floor of nModules*95/100.
As a result, we can actually have here (5 + 1/nModules) % of values which are bigger than hottest95Module.
Strictly speaking, for the 95 percentile, we obviously want the most stringent value.
It's a very pico thing but I would say we should take ceil(nModules*95/100) ?
Or did I miss sth ?

ghugo83: Suggestion (untested) : Sth like : 
irrads is a set containing doubles.
int shift =  std::ceil(nModules*95/100);  // Shift to access the 95 percentile value
 auto& hottest95Module = std::next(irrads.begin(), shift - 1);

ghugo83: Example with 2.6 mm ptPS Flat Barrel modules : 
There are 126 modules.
nModules*95./100. = 118.7 is accidentally rounded to 118.
This results in not 5.0 %, but actually 5.6% of the values in irrads being greater than the presently calculated hottest95Module.

include/AnalyzerVisitors/IrradiationPower.hh
```diff
@@ -36,7 +36,7 @@ class IrradiationPowerVisitor : public GeometryVisitor {
   std::map<ModuleRef, double> sensorsIrradiationPowerMean_;
```
alkemyst: Fixed (see change)

src/AnalyzerVisitors/IrradiationPower.cc
alkemyst: Discussed bprivately

src/AnalyzerVisitors/IrradiationPower.cc
alkemyst: The only place where the map is filled is also where a module is pushed_back into the vector and the code is a few lines above this. I am keeping this for the moment to avoid another iteration of testing.

src/AnalyzerVisitors/IrradiationPower.cc
alkemyst: As discussed privately this is correct. Adding ceil() to the code.

PR #346, opened 2017-03-30
Single hit efficiency is per-module

alkemyst:
In order to accurately study the effect of single-hit inefficiency, this was made a module-parameter

Surprisingly, this made the code run a little faster: on my laptop running with 3000 tracks on CMS-OT6.1.3-IT4.0.2.5 takes 139.7s instead of 144.5s with 95% single-hit efficiency and 112.5s instead of 117.4s for 50% single-hit efficiency.

Validation: the results are not comparable for 50% efficiency (the random effect is too large) but are compatible for 95% efficiency.

Using this patch a bug was cured: the previous parameter named efficiency was affecting the pixels too and I am not sure what efficiencyPixel was actually doing (I only noticed when trying to compare with the new behaviour). Furthermore these two parameters (which used to be in SimParms) were not designed according to the current paradigm of flexibility in properties.

### DIFF
--- a/include/DetectorModule.hh
+++ b/include/DetectorModule.hh
@@ -80,6 +80,7 @@ public:
   ReadonlyProperty<ZCorrelation, NoDefault> zCorrelation;
   ReadonlyProperty<ReadoutMode, Default> readoutMode;
   ReadonlyProperty<ReadoutType, Default> readoutType;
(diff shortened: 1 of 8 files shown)


alkemyst:
:+1:

PR #358, opened 2017-04-05
Moved all geometry classes defined on the fly in Vizard to AnalyzerVisitors/GeometricInfo

ghugo83:
This is in the process of making Vizard disappear (to be continued ..).

### DIFF
--- a/include/AnalyzerVisitors/GeometricInfo.hh
+++ b/include/AnalyzerVisitors/GeometricInfo.hh
@@ -2,9 +2,10 @@
 #define	_GEOMETRICINFO_HH
 
 #include <global_constants.hh>
-#include "VizardTools.hh"
+#include <VizardTools.hh>
 #include <Tracker.hh>
-#include <Analyzer.hh>
+#include <SimParms.hh>
+#include <TagMaker.hh>
 #include <RootWeb.hh>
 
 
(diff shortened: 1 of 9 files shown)


include/AnalyzerVisitors/GeometricInfo.hh
```diff
@@ -0,0 +1,185 @@
+#ifndef _GEOMETRICINFO_HH
+#define	_GEOMETRICINFO_HH
+
+#include <global_constants.hh>
+#include "VizardTools.hh"
+#include <Tracker.hh>
+#include <Analyzer.hh>
```
alkemyst: If the purpose is making Analyzer disappear, shouldn't we move the constants/variables/functions we need in Analyzer.hh into something different? I created AnalyzerTools for this, but any other choice is equally fine for me

alkemyst: More in general: do we need Analyzer.hh here at all?

alkemyst:
Was this PR tested?

include/AnalyzerVisitors/GeometricInfo.hh
ghugo83: Ok yes I have replaced #include <Analyzer.hh> by #include <SimParms.hh> #include <TagMaker.hh>

ghugo83:
Yes, same output ;)

PR #359, opened 2017-04-06
DevLite: Merge with tilted implementation by Gabrielle & implemented new features

drasal:
* Finished ctau calculations (taken over from tkLayout master).
* Implemented passive material type in Hit class, at which the hit has been created. Plus, a link to inactive material added. 
* Fully documented the Polygon class. Focus especially given on explanation of all mathematical concepts (intersection of polygon with a line, etc.). In addition, CoordinateOperations methods included within the Polygon class, where they naturally belong to -> CoordinateOperations class not needed anymore, hence deleted.
* User may define in SimParms, whether to use non-negligible sizes of luminous region into account, when building the detector and/or analyzing the data:
  - For details see SimParms variables: useLumiRegInGeomBuild & useLumiRegInAnalysis.
  - Luminous region defined by sigma=zErrorIP (rphiErrorIP negligible, so not taken into account). Currently uniform distr. assumed (may be easily changed in the analyzer(s) code).
  - This new approach substitutes the older approach, where zError was directly set in config file -> it may have led to inconsistent approach in different parts of detector. No such approach was supported for analysis before.
* As one shoots in analysis tracks from non-zero origin at given direction, a new checkTrackHits method was implemented for inactive elements in the same manner as for modules
  - InactiveElement classes adjusted accordingly & fully documented -> code well readable now!
  - Completely removed insur namespace (redundant in the current context)
* CheckTrackHits methods encapsulate all the functionality now: calculate intersection point & material budget. So, these quantities are calculated, where they naturally belong to (modules or inactive elements), not in analyzers.
* Several fixes in BFieldMap & IrradiationMap. Fix to Windows-like maps with special Win characters at the end of lines: \n & \r -> fix needed for correct input of units. Fix to red correctly mesh or histogram type of map: double -> long conversion introduced


drasal:
@ghugo83 Your code with tilted works now. I have merged it with lots of new features I have implemented locally during the last month. So, what do you suggest now with the tilted geometry in devLite. As far as I remember, there were some hacks, right?

ghugo83:
Heee it was already working before :) In the recent commits, I just improved the style of the Vizard code for the tilted :)
Should do for the rest of the code as well (in parallel with upstream/dev_gabie)  ;p I have started on my local but nothing finalized yet.

Hmmm maybe first thing to do would be to just test that things are running fine for older cfg files (Option1.cfg and 2.cfg) ?
They were working fine for 'flat' Option3.cfg but I don't have the 2 others.

drasal:
Hi Gabi,

it didn't work for me, that's why I mentioned that ... I've already tested that with older versions. It's fine :-)

PR #362, opened 2017-04-13
adding 50x50 configs

corrinnemills:
The only changes relative to dev_gabie should be the addition of four config files describing a pixel detector with 50x50 pixels instead of 25x100.  These are based off of OT_Tilted_362_200_Pixel_4021 rather than one of the newer layouts.

### DIFF
--- a/geometries/CMS_Phase2/OT_Tilted_362_200_Pixel_4021_5050.cfg
+++ b/geometries/CMS_Phase2/OT_Tilted_362_200_Pixel_4021_5050.cfg
@@ -1,0 +1,3 @@
+@include-std CMS_Phase2/SimParms
+@include OuterTracker/Tilted/OT_V362_200.cfg
+@include Pixel/Pixel_V4/Pixel_V4_0_2_1_5050.cfg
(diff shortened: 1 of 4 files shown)


alkemyst:
Hi Corinne,
  In order to integrate this in the main release, I'll have to ask you to follow our convention for developing and naming geometries:
  Question: why did you develop starting from 4.0.2.1? The current layout is 4.0.2.5
  Change requests:
  1) The current naming scheme is OTXXX_200_ITXXXX.cfg
  2) if you create a variant, please create a different sub-version (like 4.0.2.6 if it's based on 4.0.2.5)
  3) if you create a different subversion and you need a new included file, then this should have the same version number as the global one (e.g. in your case the pixel should be called Pixel_V4_0_2_6.cfg
  4) An entry should be added in the geometries/CMS_Phase2/readme.txt file explaining what this version is based on (quoting the previous version number) and the differences introduced.

Thanks
Stefano

alkemyst:
P.S.: I am temporarily closing this PR. Please re-open it as soon as you have fixed the naming and especially the readme.txt file
Thanks

PR #373, opened 2017-06-06
Xml export : automatization of latest updates

ghugo83:
This quick PR includes the automation of latest updates, which were used to produce the TDR XMLs.
It includes : 
- Renamed PixelBarrel to Phase2, to match new CMMSW code by Alessia in CMSSW.
- In Inner Tracker, use Phase 1 modules local frame of reference orientation, as requested by Alessia.
NB : This issue might be risen again for the Tilted Inner Tracker, as there, the Phase 2 modules local frame of reference orientation will make more sense.
- When the shape of a volume if defined, check that volume > 0.


ghugo83:
NB : This PR doesn't include the developments for the XML export of Timing Disk + Timing Barrel Layer.
These require quite much more work and will be PRed independently.

PR #379, opened 2017-06-28
OT Cabling map : Lined-up design on (-Z) side

alkemyst:


alkemyst:
Does not merge simply... will do the opposite

ghugo83:
This allows to have a lined-up Phi-nonants design: the phi sector boundaries are invariant all along Z.

The drawbacks of this design are that it implies a different cabling for disks, dependant whether the disks are placed on (-Z) or (+Z) side.

alkemyst:
Is this an optional function, or the default?

ghugo83:
The default

ghugo83:
Apparently the other design for the (-Z) side was not viable anyway, this is why I made this on 'default'.
Is that fine ?

alkemyst:
Perfectly fine, thanks

PR #403, opened 2017-11-10
Outer Tracker Services channels: MFBs and Power cables routing + Added TEDD surfaces details for technicians + Adapted cabling map to Nick TEDD design

ghugo83:
1) MFBs channels assignments:
- There should be less crossings of the MFBs than in the previous version :)
- There is the equivalent of 8 or 10 MFCs per channel.
- All double-disks within same TEDD block (TEDD_1 or TEDD_2) have same channels assignment scheme.
- There is the same channels assignment scheme on (+Z) and (-Z) sides 
('same' = invariant by push-through symmetry, mirror).

2) Power cables channels assignments:
It is now completely decoupled from MFBs channels assignments.
This scheme must follow a rotational symmetry of 180 degrees around CMS_Y.
- Barrel: There is the same cabling scheme on both (Z) sides (except for the flat ladders in TBPS).
('same' = invariant by rotation of 180 degrees around CMS_Y).
- Endcaps: 2 different cabling schemes, one per (Z) end.
All double-disks within same TEDD block (TEDD_1 or TEDD_2) have same channels assignment scheme.

3) Added detailed plots for technicians on website: which modules should be connected to which MFBs in TEDD?

4) Recomputed all cabling map for layout OT614 ('identical dees' design from Nick, resulting in a shift in Phi in TEDD).
OT614 is now set as the default layout for which the cabling map should be computed (will not work for any other layout!).

Results: 
All results accessible at http://ghugo.web.cern.ch/ghugo/layouts/channels/OT614_200_IT4025/cablingOuter.html

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -142,6 +142,7 @@ CABLING+=CablingMap
 CABLING+=DTC
 CABLING+=ModulesToBundlesConnector
 CABLING+=PhiPosition
+CABLING+=ServicesChannel
 
 EXES+=tklayout
 EXES+=setup
--- a/geometries/CMS_Phase2/OT614_200_IT4025.cfg
+++ b/geometries/CMS_Phase2/OT614_200_IT4025.cfg
@@ -1,0 +1,3 @@
+@include-std CMS_Phase2/SimParms
+@include OuterTracker/Tilted/OT_V614_200.cfg
+@include Pixel/Pixel_V4/Pixel_V4_0_2_5.cfg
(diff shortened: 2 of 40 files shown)


ghugo83:
NB: Please do not merge yet! Am creating the PR because results are stable, but am cleaning the code at the moment!

alkemyst:
Thanks for the advice.
I was still reading the comments :-)

I will then close the PR. Please feel free to re-open it as soon as you are happy with the code too.

alkemyst:
Should this be merged?

ghugo83:
Yes!

PR #404, opened 2017-11-22
Add distributions of hits in global and local frames of reference

ghugo83:
The angles in local frame of reference are the cms convention default, as defined in [1]:
reso_angles

[1] Parametrization of the spatial resolution of the reconstructed hits in the Inner Pixel for HL-LHC studies, E. Migliore and M. Musich, 2015/09/18


ghugo83:
Tested against all regressions.

ghugo83:
resolution on local y coordinate for pixel barrel modules000
Example of distribution in eta and beta, seen from BPIX modules.
All tracks come from 0 here.
One can clearly see one step in eta per BPIX layer.

alkemyst:
:+1:

PR #406, opened 2017-11-29
Fix hits on inactive elements: Effects on MB and Tracking resolution

ghugo83:
Fix InactiveElement::checkTrackHits.

The previous InactiveElement::checkTrackHits method computes several hits out of inactive volume.
The computed hitPathLength is hence sometimes huge, resulting in a non-negligible increase of the crossed MB.
This has an impact on MB and Tracking resolution.

Example on CMS Phase 2 Tracker geometry:

**Example of hit computed out of the inactive volume (before PR):**
track eta = 3.41447
trackOrigZ = -12.6505

inactive volume isVertical() = 1
inactive volume getInnerRadius() = 28.999
inactive volume getOuterRadius() = 73.557
inactive volume getZOffset() = 1103.14
inactive volume getZMax() = 1105.14

- Before PR:
hitRho = 51.242 mm
hitZ = 765.408 mm
Radiation length correction factor = 338.469

- After PR:
hitRho = 73.5508 mm
hitZ = 1104.14 mm
Radiation length correction factor = 1.00217

**Momentum resolution vs. η (const Pt across η):**

- Before PR:
before

- After PR:
after

NB: On the cfg files I have for FCC, I have no Services. 
Hence all InactiveEleement::getRadiationLength() are 0. and total inactive MB = 0.
But in case there are FCC files with inactive elements, this will have an effect!

### DIFF
--- a/source/Geometry/src/InactiveElement.cc
+++ b/source/Geometry/src/InactiveElement.cc
@@ -72,7 +72,7 @@ bool InactiveElement::checkTrackHits(const XYZVector& trackOrig, const XYZVector
   hitMaterial.radiation   = 0.;
   hitMaterial.interaction = 0.;
 
-  const double geom_zero = 1E-6;
+  const double geom_zero = 1E-6; // devLite: add in global constants
 
   // Calculate hit only if inactive element non-transparent, otherwise no effect in tracking or material budget etc.
   if (getRadiationLength() != 0. || getInteractionLength() != 0.) {
@@ -89,9 +89,9 @@ bool InactiveElement::checkTrackHits(const XYZVector& trackOrig, const XYZVector
 
 	// Geometric extrema of the inactive volume in (RZ) plane
 	const double innerZ = getZOffset();
-	const double outerZ = m_zOffset + m_zLength;
+	const double outerZ = m_zOffset + m_zLength; // devLite: add method
 	const double innerRho = getInnerRadius();
-	const double outerRho = m_rInner + m_rWidth;
+	const double outerRho = m_rInner + m_rWidth; // devLite: add method
 
 	// PART A
 	// COMPUTE THE PROJECTION OF THE TRACK IN THE 4 EXTREMA SURFACES OF THE INACTIVE VOLUME.
@@ -162,7 +162,6 @@ bool InactiveElement::checkTrackHits(const XYZVector& trackOrig, const XYZVector
 	    // Cross-check that path length is > 0.
 	    if (hitPathLength >= 0.) {
 	      // This normalizationFactor comes from the fact that MB is by default multiplied by length or width.
-	      // TO DO: Is that actually true?
 	      const double normalizationFactor = 1. / (isVertical() ? getZLength() : getRWidth());
 	      // Compute MB
 	      hitMaterial.radiation   = getRadiationLength() * normalizationFactor * hitPathLength;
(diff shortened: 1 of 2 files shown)


drasal:
Hi Gabi,

interesting! Let me check the impacts on the FCC geometry before merge (will do it early next week)...

drasal:
Hi Gabi,

I have a question, I somehow didn't get how the algorithm works with a horizontal inactive volume, when shooting e.g. at eta>0, but origZ=-somevalue. I somehow expected different approach when dealing with horizontal versus vertical inactive element, but I may be wrong... 

Then I have another question, regarding line 167:

getRadiationLength() * normalizationFactor * hitPathLength;

The radiation length is expressed per unit length, so one needs to scale it by path length, right? But why another normalization factor. The path length should already reflect whether it's horizontal or vertical shape. But I may be wrong, this may be connected with my previous question...

drasal:
Otherwise, the algorithm is really needed, the previous version was clearly wrong! So thanks a lot for addressing that....

drasal:
Gabi,
one last comment, the geom_zero variable is to find out whether track perpendicular or parallel to Z, correct? If yes, one could in principal get the same information directly from direction.Eta() or direction.Theta()... But that nothing serious...

ghugo83:
Hi Zbynek!

To answer your points:

1) Point on MaterialProperties::getRadiationLength().
Yes I was surprised about it as well, but MaterialProperties::getRadiationLength() (and same for getInteractionLength() ) does not return radiation length per unit of length.
The value returned by MaterialProperties::getRadiationLength() is already multiplied by distance d:
- for a tub: d = width.
- for a ring: d = length.
I guess the assumption of whoever developed that, was that d would be the length of the track crossing the inactive volume.
As you know, it is in the end not sharp correct. Hence, one needs to divide by d, and multiply by the actual track length through the Inactive Volume.

This is not a bug (well, apart that it is inelegant to multiply by d as the first place!), and was not touched by my version not yours.
You have this 1/d factor in your code as well: it is just that you incorporated it in trackPathLength, while I made it explicit.

So there is no difference between old code, devLite, or this PR on this point :)

ghugo83:
2) Different approach for horizontal or vertical volume.

I deducted that this distinction was at first because of this 1/d factor.
Then, I guess people progressively added code on top of that split.

But, if you think of it, geometrically speaking, there is no reason to separate tub from ring.
Since we are already (devLite and this PR) taking all cases into account, one just need to look through which side(s) the track is crossing the inactive volume :)

This makes things much lighter :)

ghugo83:
geom_zero is just there to avoid rounding issues around 0. :
-1E-17 to be considered different from 1E14, or similar :)

ghugo83:
You are welcome :)
Yes this PR solves the bug that was observed, and also makes things more compact :)
Bonne soiree :)

drasal:
Hi Gabi,

thanks  a lot for all the explanations. I agree with all these and actually I didn't expect that it would be wrong (you do intensive calculations before implementation), I just wanted to understand. In addition to all this, can you explain to me what the 4 cases stand for, what do you mean by:
* Track enters and exits the volume between extrema radii cylinders.
* Top left corner.
* Bottom right corner.
* Track enters and exists the volume between extrema Z planes.

Merci beaucoup et bonne soiree, Madame :-)

ghugo83:
Yes sure Zby (can I call u like that?) :)

The 4 cases are the following:
4_cases
You have this in your code as well. But as you can see, there is no need to distinguish tub or ring here, since everything is already taken into account.

Please note that the effect of the 'top left' or 'bottom right' cases (already in devLite) on the total MB are negligible though.

The effect of the PR is that the hit is always properly computed now (before PR, the computation was making the hit sometimes computed out of the inactive volume, and resulting computed path was hence sometimes huge).

Bonne soiree :)
Cheers

PR #409, opened 2017-12-01
Merge PR#332 from Zbynek: renaming HitNew -> Hit, TrackNew -> Track

ghugo83:
This is the merge of https://github.com/tkLayout/tkLayout/pull/332 .
All regressions corrected :)
All merging conflicts solved :)

**This PR is improving style, there is no effect on the results.**

Before PR:
http://ghugo.web.cern.ch/ghugo/layouts/test/OT613_200_IT404_patternreco_70luminous_triggerfixedDistance/info.html 

After PR:
http://ghugo.web.cern.ch/ghugo/layouts/test/OT613_200_IT404_merge/info.html

PR #405, opened 2017-11-29
Important: Bug fixes affecting tracking resolution + Shape of luminous region for resolution studies

ghugo83:
**In this PR:**
- A) Instead of IP = (0,0,0), a flat luminous region from Z = -70 mm to Z = 70 mm is chosen for resolution studies.
- B) BUG FIX:  One should not select the Z > 0 inactive volumes only (assuming eta >= 0). Indeed, the IP can be at Z < 0, hence inactive volumes on (-Z) side can be hit.
- C) BUG FIX:  One should not select the Z > 0 sensors only (assuming eta >= 0). Indeed, the IP can be at Z < 0, hence sensors on (-Z) side can be hit.
- D) BUG FIX: Sensors tilt angle was not taken into account for crossed MB calculation, in tracking resolution studies.
- E) Improved code on hits on inactive elements. This is used to calculate hit position and crossed MB. (though, on this point, issues were rather in devLite than in master).

**Results (on OT613_200_IT404 geometry) :**
- Before this PR, with Ernesto's parametrization: http://ghugo.web.cern.ch/ghugo/layouts/July/OT613_200_IT404/errorstracker.html 
- Before this PR, with Riccardo's parametrization: http://ghugo.web.cern.ch/ghugo/layouts/September/OT613_200_IT404/errorstracker.html
- After merging this PR, with Riccardo's parametrization: http://ghugo.web.cern.ch/ghugo/layouts/channels/OT613_200_IT404/errorstracker.html 

The effects are significant on resolution (pixel), resolution (tracker) and resolution (trigger).
**Example: Longitudinal impact parameter resolution vs. η (const Pt across η):**
- Before this PR: 
before

- After merging this PR: 
after

**NB: Details:**
A detail of the impact of the fixes can be found here:
- Initial: http://ghugo.web.cern.ch/ghugo/layouts/test/OT613_200_IT404_distr_InactiveEtaMinMaxFixed_zError0/errorstracker.html 
- +A: http://ghugo.web.cern.ch/ghugo/layouts/test/OT613_200_IT404_distr_InactiveEtaMinMaxFixed/errorstracker.html
- +B: http://ghugo.web.cern.ch/ghugo/layouts/test/OT613_200_IT404_distr_InactiveEtaMinMaxFixed_NoMax/errorstracker.html 
- +C: http://ghugo.web.cern.ch/ghugo/layouts/test/OT613_200_IT404_distr_InactiveEtaMinMaxFixed_NoMaxMod/errorstracker.html
[...]

PR #410, opened 2017-12-04
Shape of luminous region directly specified in SimParms 

ghugo83:
The shape of the luminous region can now be directly specified in SimParms, instead of being hardcoded in Analyzer.

Present choice is IP = (0, 0, U[-70, 70]) for all analysis, except MB analysis which uses IP = (0,0,0).

NB: Shape of luminous region for trigger studies is switched here from Gaussian(0,70) to U[-70, 70].


ghugo83:
IMPORTANT NB:
The PRs to dev_gabie should be merged in the chronological order:

    PR #405 (bug fixes in tracking resolution + pattern reco)
    PR #407 (bug fix in trigger efficiency results)
    PR #408 (extended shape of luminous region in pattern reco)
    PR #409 (code style)
    PR #410 (shape of luminous region in SimParms)

Each time I create a new PR, I keep the previous commits from previous PRs, so that all previous changes and fixes are incorporated, and so that there will be no merging conflict.

PR #451, opened 2018-08-30
Routing of services: allow to go towards smaller |Z|

ghugo83:
So far, the only supporting direction is to go towards bigger |Z| and bigger |R|.

Going towards smaller |Z| was needed for placing the conversion stations over TBPX.
The services exit TBPX at (barrel's maxZ, barrel's  maxR). 
The services can now go all the way to the conversion station over TBPX. 
Then, after meeting the conversion station, they come back towards bigger |Z|.

tbpx_conversion_station_and_services_routing

The logics is the following:
- Compare barrel's maxZ with the conversion station meanZ. 
If station's meanZ <  barrel's maxZ, the 'C-shape services' mode is launched.

- 'C-shape services mode': build 2 extra sections.
One cabling section goes from barrel's maxZ to the conversion station. 
The other section goes back from conversion station to barrel's maxZ, and then plug itself to the existing routing.

- Microscopic point of view: needed to add support in buildSection and splitSection functions, to be able to go towards smaller |Z|.
The entire implementations of these functions were assuming to go towards bigger |Z|, and were not valid here. 
Additionally, added support for going towards smaller |R| (not needed but never knows).

Off-topic: This PR also includes:
- A fix on the conversion stations error message. 
This was a complaint on conversion stations which were already built, while trying to build them again, which should not happen.

- A fix on the info/warning/error color code on website.
Before PR: yellow = error, orange = warning, red = info.
After PR: yellow = info, orange = warning, red = error.
Now that the errors on conversion stations are fixed, we are at the warning (orange) level.

- Remove unused constants.

NB:  geometries/CMS_Phase2/Pixel/Pixel_V4/stations_BPIX_over_TBPX will eventually be deleted
(4 stations over TBPX instead of 1).
It is temporary introduced for debugging purposes.


ghugo83:

PR #452, opened 2018-08-30
Also add IT 460 and IT 461 cfg files.

ghugo83:


ghugo83:
Forgot to add these 2 files.

PR #453, opened 2018-08-31
Fixed e-links -> optical fibers conversions in materials cfg files

ghugo83:
The numbers of GBTs in TBPX are:
L1: 108
L2: 96
L3: 80
L4: 56
The colours of the TBPX conversion stations are directly proportional to the number of GBTs (GBT-DCDC file is the only local material in the stations).

- Fix A:  
**The colors of the TBPX conversion stations should show progressively lighter MB from L1 -> L4.**

This was not the case:
rainbow_before

This was due to 1 e-link/ module counted for L3 instead of 2 e-links / module, in the materials cfg files.
Hence the number of GBTs in L3 was not correct.
After fix:
rainbow_after

- Fix B:  
**If one scales the length of the stations proportional to the number of GBTs, all stations should be of same colour.**
This was not the case:
4stations_before

This is due to approximations in the number of modules / GBT in the cfg material files: one could use a non-integer value, and could use a different file for each TBPX layer.
L1: 1 module / GBT -> 1 module / GBT
L2: 3 modules / GBT -> 2.25 modules / GBT
L3: 3 modules / GBT -> 2.25 modules / GBT
L4: 6 modules / GBT -> 4.5 modules / GBT

After fix:
4stations_after

NB: TO DO: Could use a dedicated cfg file for each TFPX and TEPX ring as well, then remove the TWP cfg files.


ghugo83:
Overall, the diff in MB is:
- before PR: http://ghugo.web.cern.ch/ghugo/layouts/elinks/OT614_200_IT404_rainbow_before/materialpixel.html 

- after PR: http://ghugo.web.cern.ch/ghugo/layouts/elinks/OT614_200_IT404/materialpixel.html 

The diff in overall RL for GBTs + DCDC materials, as well as for optical fibers, is ~10% to 13% increase.

PR #455, opened 2018-09-03
Mini-PR: Adjusted Z positions of BPX conversion stations on service cylinder in IT461

ghugo83:
BPX conversion stations on service cylinder are now nearly adjacent, from Z = 430 mm to Z = 820 mm. Since they are 2 mm thick only, making them shorter result in a huge MB concentration in one spot only. 
To be further adjusted by Andre.


ghugo83:
Using the present tkLayout/dev (after merge of this PR), a Radiation Length comparison between IT 460 and IT 461 is accessible here:
tklayout_sim_mb_it460_versus_it461

PR #460, opened 2018-09-27
Add volumes to Pixel Modules

ghugo83:
Description at https://indico.cern.ch/event/750045/contributions/3150750/attachments/1720535/2777743/news_from_materials_modeling.pdf 

NB: dev branch is not fully debugged & cleaned AND SHOULD NOT be merged into master.
This is merged into dev in order not to block the cfg files descriptions improvements.


ghugo83:
comp
Comparison CMSSW / tkLayout with the new materials model show no major inconsistency : uncertainties are in the same order of magnitude as before the change of materials model.

Warning: I need to do a deeper cleaning and debug though.

PR #463, opened 2018-10-15
Avoid TCanvas::DrawClone() 

ghugo83:
There was a heavy and unnecessary use of TCanvas::DrawClone().
This was is order to avoid transferring TCanvas objects ownership from Analyzer & Vizard to RootWeb.
This was considerably increasing run time.

Since TCanvas has no move semantics, just use unique_ptr<TCanvas> in the caller code instead!!


ghugo83:

PR #466, opened 2018-11-01
Adds support of new rings paradigm in TEPX

ghugo83:
bigDelta: distance in Z between the center of a disk and the center of a ring.
smallDelta: distance in Z between the center of a ring and the center of a module.

Adds support for smallDelta > bigDelta case.
Obviously still need to keep support for smallDelta < bigDelta.

smallDelta > bigDelta: Modules of a ring are placed on 4 dees.
smallDelta < bigDelta: Modules of a ring are placed on 2 dees: half of the modules on each dee side.

See page 4 at https://indico.cern.ch/event/746593/contributions/3164208/attachments/1726785/2789605/EPIX_Geometry_02102018.pdf.

In practice, what needed to be adapted was:
- Added info on Ring and EndcapModule classes about how the configuration is.
Ring::isRingOn4Dees() (is smallDelta > bigDelta ?)
Ring::isSmallerAbsZRingInDisk() (is ring placed at -bigDelta in its disk ?)
EndcapModule::isSmallerAbsZModuleInRing() (is modules placed at -smallDelta in its ring)?

- Adapted modules flip.

- Adapted IT cabling map.

- Adapted XML export.

Also add IT layout IT 612, with smallDelta = 4 mm and bigDelta = 2 mm.

### DIFF
--- a/geometries/CMS_Phase2/OT614_200_IT612.cfg
+++ b/geometries/CMS_Phase2/OT614_200_IT612.cfg
@@ -1,0 +1,3 @@
+@include-std CMS_Phase2/SimParms
+@include OuterTracker/Tilted/OT_V614_200.cfg
+@include Pixel/Pixel_V6/Pixel_V6_1_2.cfg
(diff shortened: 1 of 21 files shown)


ghugo83:

PR #478, opened 2019-02-14
Merged friday's code: Assign MFB fanout branches to modules in TEDD

ghugo83:
OT cabling map, TEDD: 
Each MFB has 4 fanout branches, for the connections to modules.

The logics used to be simple: for each MFB, 1 fanout branch per disk surface.
Now, there is the need, for each MFB, in case the modules of a given disk (not double-disk!) belong to different dees, to make an assignment per dee instead.

This is fully automated, so that the Mechanics can pick up whatever is in AggregationsPatterns.csv

All results are at: http://ghugo.web.cern.ch/ghugo/layouts/cabling/OT616_200_IT404/cablingOuter.html 
One can visualize the connections at the 'modules to bundles' section on the webpage. 
For example, if you look at the 'Positive cabling side', you can look at the 4 surfaces of DD 1, TEDD1.
The alternation of colors show the alternation of MFBs the modules end up connected to. The color scheme is not unique.
For a given MFB (one color), the alternation of transparent / non-transparent show the alternation of the fanout input slots that are used.

### DIFF
--- a/include/DetectorModule.hh
+++ b/include/DetectorModule.hh
@@ -375,15 +375,16 @@ int numSegmentsEstimate() const { return sensors().front().numSegmentsEstimate()
   std::string summaryFullType() const;
 
   // OT CABLING
-  void setBundle(OuterBundle* bundle) { bundle_ = bundle ; }
+  const int isPositiveCablingSide() const;                    // Can be different from (Z) end. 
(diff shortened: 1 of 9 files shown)


ghugo83:
Checked against all regressions
To come: probably need to add info for (-Z) end.

PR #480, opened 2019-02-15
Adds support for bigParity on tilted rods phi placement. 

ghugo83:
This allows to change design. It comes from a misinterpretation from the Mechanics, that would be hard to correct on their side. 
This change in tracker design should obviously have no impact on tracker performance.


ghugo83:
Before PR:
img003

After PR:
img004

ghugo83:

PR #486, opened 2019-03-07
Build system improvements and c++17 compatibility

vvolkl:
A lot of custom cmake code can be replaced by standard calls to `find_package`. The change in `Materialway.hh` is necessary to compile tklayout with c++17.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -59,6 +59,7 @@ INCLUDE_DIRECTORIES( ${PROJECT_SOURCE_DIR}/source/Geometry/include
                      ${PROJECT_SOURCE_DIR}/source/Tools/include
                      ${PROJECT_SOURCE_DIR}/source/Tracking/include
                      ${PROJECT_SOURCE_DIR}/source/Web/include
+                     ${PROJECT_SOURCE_DIR}/include
                      ${Boost_INCLUDE_DIR}
                      ${ROOT_INCLUDE_DIRS} )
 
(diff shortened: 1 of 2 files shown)


ghugo83:
Thanks a lot @vvolkl for this contribution!
Are you sure this is intended to master branch? I tried and it does not compile on master branch.
- pedantic is not supported yet by the code itself.
- instead, -Wno-reorder is not needed.
- there is no _source_ repo on master branch, only _include_ and src.

vvolkl:
Hi @ghugo83,
it's true, I originally did these changes for the masterLite branch, I may have forgotten something. Also, I did not actually use `setupCMake_slc6.sh` to compile but the FCC software environment. I'll give it another try!

ghugo83:
Hi @vvolkl 

In master and dev the environment (and hence tkLayout codebase) needed quite some modernization first.
I have done that in https://github.com/tkLayout/tkLayout/pull/487 

Could you please rebase your PR (#486), to the tkLayout/dev branch, if you have some time?
It would be very beneficial to have the cleaning you did of all the verbosity in CMakeList.txt also available in tkLayout/dev :)
Thanks a lot for this :)

NB 1: Evth works with the new Makefile.
Instead, CMakeList.txt needs to be updated :) Still C++11 in CMakeList.txt ;)

NB 2: tkLayout/dev will eventually be merged into tkLayout/master, but at a later stage (for other reasons).

vvolkl:
Great, seems like the compilers and dependency versions are really up to date now. Sure,I rebased on dev  in #488 (For the setup script it is easier with cmake to just use the LCG view setup than setup individual packages  so I changed that also)

ghugo83:
Thanks a lot!
Regarding the LCG view: I independently setup the 3 packages we rely on, instead of the full LCG view, so that the dependencies are explicit and minimal (otherwise one looses track of what we rely on).
I will close this PR as you adapted it to dev in https://github.com/tkLayout/tkLayout/pull/488

PR #487, opened 2019-03-15
Port project to modern compilers (gcc 4.7.2 to 8.2.0) and C++17 (well, obviously only support for C++17 which is already implemented in gcc 8.2.0)

ghugo83:
Port environment to some modernity:
* gcc 4.7.2 -> gcc 8.2.0
* ROOT 5.34.09 -> ROOT 6.16.00
* BOOST 1.50.0 -> BOOST 1.69.0
* doxygen 1.8.2 -> 1.8.11

Compiler flag:
* std=c++11 -> std=c++17.

WARNING: Makefile was updated but NOT CMakeLists.txt

NB 1: Cheat intro to LCG: https://twiki.cern.ch/twiki/bin/view/Persistency/PersistencyCMake 

NB 2: Obviously not all C++17 features are supported by gcc 8.2.0.
Nice presentation on which C++ feature is supported by which compiler on https://gcc.gnu.org/projects/cxx-status.html


ghugo83:
Ckecked against all regressions.
Only one diff still remaining: the plotting of the TH1D regarding local spatial resolution is different but ok.

PR #488, opened 2019-03-15
CMake CleanUp

vvolkl:
Update of #486

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,18 +14,12 @@ CMAKE_MINIMUM_REQUIRED(VERSION 2.6 FATAL_ERROR)
 # Project name and version
 PROJECT(tklayout)
 
-#set( ${PROJECT_NAME}_MAJOR_VERSION 1 )
-#set( ${PROJECT_NAME}_MINOR_VERSION 0 )
-
 #----------------------------------------------------------------------------
 # Project options
-OPTION( INSTALL_DOC "Set to OFF to skip build/install Documentation" ON )
+OPTION(INSTALL_DOC "Set to ON to build/install Documentation" OFF )
 
-#----------------------------------------------------------------------------
-# Advanced settings
-# Library *nix style versioning
-SET( ${PROJECT_NAME}_SOVERSION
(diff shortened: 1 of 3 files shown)


ghugo83:
@vvolkl Thanks a lot!
Please not the const return value in the comparator in Materialway.hh is not strictly needed for successful compilation in C++17 (only the comparator itself needs to be const). 

Am testing this PR now ;)

ghugo83:
Hmm it does not seem to find boost properly :(
Did you test it? Could you test & update?

vvolkl:
Hi @ghugo83 
ah, right, I was talking about changing the `setupCMake_slc6.sh` but then forgot to push it. Since you anyway prefer not sourcing the whole view I updated it in a different way using the `CMAKE_PREFIX_PATH`. I had to also use a newer cmake to use the `CMAKE_CXX_STANDARD` variable. There are some cmake warnings about boost, but it should be ok nonetheless! Could you give it another try?

vvolkl:
Oh, and it seems it's true that the change in the comparator is not needed (anymore? at one point I ran into a compilation error and had to do that to fix it, so there might be some issue with a different compiler). so I removed it.

ghugo83:
Thanks a lot @vvolkl !
- The cmake warnings about boost are due to version of cmake which is still slightly too old.
Loading /cvmfs/sft.cern.ch/lcg/contrib/CMake/3.13.4 instead of 3.8.1 fixes it (otherwise, cmake version is older than boost one).
This also fixes a warning at run time.

- What about compiler flags? 
-std=c++17 -Wall -Werror -fmax-errors=2 -g 
are presently in Makefile, would be nice to have the same flags here.
I do not think -Wall is there by default for example.
(actually would be nice to have more flags but that's a full other story :))

Once these 2 points are fixed I ll merge ur PR :)
Thanks a lot :)

vvolkl:
Good catch, the newer cmake indeed gets rid of the boost warnings. The "-std=c++17" is already set up by cmake via `CMAKE_CXX_STANDARD`, the other flags I added in addition.

No problem! I think FCC profited a lot from tkLayout so it's nice to contribute something (even if it's just a small patch) :)

ghugo83:
Thanks @vvolkl , I am merging the cleaned CMakeLists now ;)
Cheers

PR #496, opened 2019-05-23
bugfix: it would not compile!

alkemyst:


ghugo83:
Yes I fixed it in PR #495 
but now there is merge conflict :/

PR #504, opened 2019-07-17
Clean

mehendibasu:


ghugo83:
There was an issue with the skewAngle.
Also needed to be more generic to be able to flip bigParity.
Fully finalized in https://github.com/tkLayout/tkLayout/pull/505

PR #512, opened 2019-11-18
Riccardo s new parametrization: lower threshold: 1500 e- -> 1000 e-

ghugo83:
No major diff, reso slightly improved (since lower threshold).

IMPORTANT: param for 50x200 not updated.

Before PR:
- http://ghugo.web.cern.ch/ghugo/layouts/test/OT616_IT614_old_reso/layoutpixel.html 
After PR:
- http://ghugo.web.cern.ch/ghugo/layouts/test/OT616_IT614_new_reso/layoutpixel.html

PR #517, opened 2020-01-08
Riccardo s new parametrization: minor correction in 25 um parametrization.

ghugo83:
Recent update in parametrization: Lower threshold: 1500 e- -> 1000 e-
The shape at normal incidence (when lorentz angle = 0) was a hole, which was not expected.


ghugo83:
See PR #512

ghugo83:
Validated behavior at normal incidence (when B = 0).

**Barrel:**
Before:
before_barrel_x

After:
after_barrel_x


**Endcaps:**
Before:
before_endcap_x

After:
after_endcap_x

ghugo83:
Effect on pT resolution, d0 resolution, phi resolution is visible but not major (slight degradation).

PR #520, opened 2020-01-22
Add OT 617 (and OT 618): Increase TEDD outermost radius to improve transition with TB2S (and TBPS)

ghugo83:
- All TEDD: update smallDeltas to latest info.
PS 4 mm: 7.375 mm -> 7.550 mm.
2S 4 mm:  8.550 mm -> 8.595 mm.
2S 1.8 mm: 7.450 mm -> 7.495 mm.

- Improve transition region between TB2S and TEDD by removing readout hybrid in TEDD1, R15.
Special modules in TEDD1, R15, with halved number of strips / module. WARNING: MODULE MB NOT UPDATED!!
Outer radius increased by 20.47 mm : R15 sensors centers: 1023.16 mm -> 1043.63 mm.
Radii in TEDD1, R12, R13, R14 adjusted accordingly.

- OT618: do the same thing on TEDD 2 R15 as well.

- The idea behind this is to reduce the gap at eta ~ 1.05.
OT 617 layout might never be built, if the performance is not shown to improve significantly.
OT 618 layout is not likely to be built, since TEDD 2 design (Lyon) is already frozen.

### DIFF
--- a/config/stdinclude/CMS_Phase2/OuterTracker/Materials/ModuleMaterials/Components/2S_1Strip_referenceSensors
+++ b/config/stdinclude/CMS_Phase2/OuterTracker/Materials/ModuleMaterials/Components/2S_1Strip_referenceSensors
@@ -1,0 +1,11 @@
+// 2S, 1 Strip, reference sensors
+
+// Default sensor:
+ReferenceSensor 1 {
+  numStripsAcross 1016
+  numSegments 1
+}
+ReferenceSensor 2 {
+  numStripsAcross 1016
+  numSegments 1
+} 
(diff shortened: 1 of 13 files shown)


ghugo83:
Effect on performance:

- ref: http://ghugo.web.cern.ch/ghugo/layouts/test/OT616_IT613_fix_pT/errorstracker.html 
ref

- OT 617: http://ghugo.web.cern.ch/ghugo/layouts/test/OT617_IT613_test/errorstracker.html
OT_617

- OT 618: http://ghugo.web.cern.ch/ghugo/layouts/test/OT618_IT613/errorstracker.html
OT_618

PR #521, opened 2020-01-22
Fix binning in info + geometry + tracking resolution plots

ghugo83:
- deltaEta : 0.1 -> 0.02 on all plots.
- removed spot markers, which are hiding the interesting information.

This allows to see (hence understand...) way better the geometric coverage, as well as the details of the tracking resolution.

NB: deltaEta has 2 exceptions:
- deltaEta = 0.04 in 'Distribution of number of stubs per track' plots.
- deltaEta = 0.01 in 'Number of layers with at least a hit' plots. Might be nice to have it as well for the 'Number of hits' plots.


ghugo83:
Before PR:
http://cms-tklayout.web.cern.ch/cms-tklayout/layouts-work/recent-layouts/OT616_IT613/info.html 

After PR:
http://ghugo.web.cern.ch/ghugo/layouts/test/OT616_IT613_bining/info.html

PR #530, opened 2020-02-14
Export to CMSSW: Automatic export of OTST, with no clash with BTL

ghugo83:
Since introduction of MTD:
- the OTST is placed directly inside the Tracker volume.
- the OTST is described in a dedicated file (otst.xml).

Until now, OTST stayed static in CMSSW, and did not benefit from the updates from Mechanical engineers, which were studied and debugged in tkLayout.

As was in the to-do list for a long time:
**create an automatic export of the OTST.**
The difficulty lied in that the OTST needs to be seen as a different subdetector by tkLayout XML export, because it needs to be placed in its dedicated file, with its dedicated DD sections.

This allows to introduce the latest description of the OTST from Mechanics:
- CFRP skin thickness: 2 mm -> 4 mm. This explains a factor ~2 in the OTST weight!
- CFRP skin density: 1690 kg/m^3 -> 1550 kg/m^3.
- half length is updated from 2900 mm to 2650 mm.
- addition of Core materials (HexWeb HRH-10-1/4-3.1).

The effect on MB in CMSSW is expected to be significant (~ +150 kg), but obviously outside of Tracking volume.

### DIFF
--- a/include/Extractor.hh
+++ b/include/Extractor.hh
@@ -85,7 +85,7 @@ namespace insur {
       }
     };
   public:
-    void analyse(MaterialTable& mt, MaterialBudget& mb, XmlTags& trackerXmlTags, CMSSWBundle& d, bool wt = false);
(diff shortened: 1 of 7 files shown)


ghugo83:
**Effect on full OT volume, RL:**

otst_ratio

PR #535, opened 2020-06-08
Inner cabling map: Created CMSSW Id schemes for GBTs and DTCs (cache friendly for runs on GPUs)

ghugo83:
Created CMSSW GBT and DTC Id schemes, not convenient for debugging, but better for the cabling map (in CMSSW) to be run on GPUs.

- GBTs CMSSW Ids are consecutive integers, and they are Ids PER DTC (from 1 to 60 max).
- DTCs are consecutive integers (from 1 to 28).

After PR: http://ghugo.web.cern.ch/ghugo/layouts/cabling/OT616_IT613_CMSSW_Ids/cablingInner.html

### DIFF
--- a/include/InnerCabling/GBT.hh
+++ b/include/InnerCabling/GBT.hh
@@ -48,8 +48,9 @@ public:
 
   // GENERAL INFO ON THE GBT
   const std::string GBTId() const { return myGBTId_; }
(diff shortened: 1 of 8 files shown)


ghugo83:
Fully debugged.

PR #555, opened 2021-02-05
Include latest TFPX changes

adewit:
@alkemyst this is what is currently being used for the latest TFPX layout. Marking this as a draft PR because there are a few things that would be really nice to improve (in parallel with the various CMSSW studies that are needed)
- Set r/twist with a vector in the config (like it is now done in the CMSSW export); the current implementation makes some changes to the underlying property method that are ugly and undesirable
-  For the xml export, some vectors that aren't part of the ringInfo struct are used. Again, this works for now, but it would be better to find a way to include them in the struct.
(Some of my own debug printout needs to be removed, too :/)


include/DetectorModule.hh
```diff
@@ -224,6 +230,10 @@ public:
 
   const XYZVector& center() const { return decorated().center(); }
   const XYZVector& normal() const { return decorated().normal(); }
+  const XYZVector& cornerone() const { return decorated().cornerone(); }
+  const XYZVector& cornertwo() const { return decorated().cornertwo(); }
+  const XYZVector& cornerthree() const { return decorated().cornerthree(); }
+  const XYZVector& cornerfour() const { return decorated().cornerfour(); }
```
alkemyst, line 236: Is this a function to extract the module's corners? I thought it already existed.

include/DetectorModule.hh
```diff
@@ -247,6 +257,8 @@ public:
   const bool isAtPlusXSide() const { return (center().X() >= -insur::geom_zero); }
   double tiltAngle() const { return tiltAngle_; }
   bool isTilted() const { return tiltAngle_ != 0.; }
+  double twistAngle() const { return twistAngle_; }
+  const XYZVector& getRAxis() const {return rAxis_;}
```
alkemyst, line 261: Is this the centre of the sensor? Again, I thought this vector is already available.

include/Ring.hh
```diff
@@ -54,6 +55,7 @@ class TiltedRing : public PropertyObject, public Buildable, public Identifiable<
 
  TiltedRing() :
   //materialObject_(MaterialObject::ROD),
+    moduleNode            ("Module"                , parsedOnly()),
```
alkemyst, line 58: Out of my ignorance: why do you need to add Module here?

alkemyst: Ah, o: I get it: to assign individual rhoShifts and twists, is that correct?

include/tk2CMSSW_datatypes.hh
```diff
@@ -263,6 +263,7 @@ namespace insur {
         std::string childname;
         bool isDiskAtPlusZEnd;
         int numModules;
+        int discNumber;
```
alkemyst, line 266: Why do you need `discNumber` here? We usually use `layerNumber` to count discs in the forward

src/Ring.cc
```diff
@@ -79,13 +79,45 @@ std::pair<double, int> Ring::computeOptimalRingParametersRectangle(double module
...
+  double deltaPhiLarge = deltaPhiNom+2*phiShift;
+  double nominalZRot = 0.;
+  if (deltaPhiLarge!=deltaPhiNom){ deltaPhiNom = deltaPhiNom-(4*phiShift/(numMods-2));}
   for (int i = 0, parity = smallParity(); i < numMods; i++, parity *= -1) {
     EndcapModule* mod = GeometryFactory::clone(*templ);
+    if (moduleNode.count(i) > 0){ mod->store(moduleNode.at(i));}
     mod->myid(i+1);
-    mod->rotateZ(2.*M_PI*(i+alignmentRotation)/numMods); // CUIDADO had a rotation offset of PI/2
+    std::cout<<"------begin module printout--------"<<std::endl;
+    if(mod->propertyTree().get<double>("twistAng",-99) > -99 ){//Need to do this before further determining phi/z of the module
```
alkemyst, line 93: I would rather choose -999which is not a valid angle

Actually at a point we should consider switching these return values to using std::optional<>, but that would require a lot of work...

include/DetectorModule.hh
ghugo83: Additionally: please use camel case notation.

include/DetectorModule.hh
ghugo83: Would be nicer to call it yawAngle (not important, but that makes the code more readable eventually because one can see out of context the type of rotation the variable refers to)

include/DetectorModule.hh
```diff
@@ -146,6 +147,9 @@ public:
 
   Property<bool, Default> removeModule;
 
+  Property<double, Default> twistAng;
```
ghugo83, line 150: yawAngle
Do not hesitate to not abbreviate the names, it makes the reading easier, and eventually the most time is taken understanding the code, while typing is an infinitesimal part.

include/DetectorModule.hh
```diff
@@ -281,11 +293,13 @@ public:
   void rotateToNegativeZSide() {
     side(-side());
     rotateY(M_PI);  // Rotation around CMS_Y of angle Pi
+    twistAngle_=-twistAngle_; // Flip twistAngle to match how the modules are actually placed
```
ghugo83, line 296: yawAngle*=(-1.);

include/GeometricModule.hh
```diff
@@ -72,6 +77,7 @@ public:
  
   double tiltAngle() const { return tiltAngle_; }
   double skewAngle() const { return skewAngle_; }
+  double zrotAngle() const { return zrotAngle_; }
```
ghugo83, line 80: camelCase
yawAngle

geometries/CMS_Phase2/Pixel/Pixel_V7/FPIX1_7_1_0.cfg
```diff
@@ -0,0 +1,454 @@
...
+       twistAng -0.148353
+       rhoCentre 53.31
+      }
+      Module 1 {
+       twistAng -0.1153663
+       rhoCentre 53.08
+      }
+      Module 2 {
+       twistAng -0.0349066
+       rhoCentre 52.76
```
ghugo83, line 42: I would have expected 3 variables here:
- Xshift         // translation of sensor center of reference in X on the dee plane
- Yshift         // translation of sensor center of reference in Y on the dee plane
- yawAngle  // rotation of the sensor around its center, on the dee surface
Does that mean that all sensors are equidistant in phi?

In any case, maybe does not harm to have the 3 degrees of liberty, so that any further adjustment can be integrated?

Please note that the precision can be more limited using angles than distances at the CAD -> cfg files interface.
For yawAngle we have no choice, but for the sensor translation, a translation can be good.

src/Extractor.cc
```diff
@@ -2168,14 +2168,53 @@ namespace insur {
...
-	//if (zmin > 0) {	
+	//if (zmin > 0) 	
         std::ostringstream dname, pconverter;
 	//disk name
         dname << xml_disc << discNumber; // e.g. Disc6
 
         std::map<int, ERingInfo> rinfo;
 	std::set<int> ridx;
 
+
```
ghugo83, line 2179: Yes I had a glance at your presentation's slides last Friday.
The Ring algo is used to place modules within a ring, in, for example, the forward section, in order to avoid lengthy XMLs.
Basically, an other option would have been to place each module individually within a ring (ie, numModulesPerRing XML Pospart sections).
Since the ring was regular, there was no point doing so.

Now, since you need to have specfic position for each module within a ring, there is no point using the algorithm. It will not work without modifying the algo in CMSSW, with risks of regressions in all the numerous placed it is used, and additonnally has no advantage, because there is no regularity that needs to be abstracted into an algo here.

When isRegularRing() is wrong, you could just remove the algo section, and instead add numModulesPerRing sections, adding each single module position within the Ring.
You could simply describe these PosPart placements with one translation and one rotation (which is exactly what you need).

src/Extractor.cc
```diff
@@ -2168,14 +2168,53 @@ namespace insur {
...
+      
+        std::map<int, std::vector<double>> phi_one;
+        std::map<int, std::vector<double>> phi_two;
+        std::map<int, std::vector<double>> radius_one;
+        std::map<int, std::vector<double>> radius_two;
+        std::map<int, std::vector<double>> twist_one;
+        std::map<int, std::vector<double>> twist_two;
+
+        for (iiter = oiter->begin(); iiter != oiter->end(); iiter++){
+          if(iiter->getModule().uniRef().subdetectorName=="FPIX_1"){ //Don't need to check this for every endcap ring - awful hack, to be improved
```
ghugo83, line 2188: Yes please never use harcoded subdetector names. Otherwise the codebase become a CMSSW-style mess ;p
The only place where there are harcoded names is in the cabling (in a dedicated global string header), so that it is launched only on one given layout name (the cabling is layout-specific).

Here, you need to see what is the feature that makes the ring treatment different, and typically add a member variable in the ring hierarchy level, to acknowledge that.
You could add in the cfg files, when the irregualr mode is used, a bool:
isRegularRing false;
Which will set a boolean data member within the Ring class at CONSTRUCTION time. 
Then a getter in the Ring class:
isRegularRing() const { return isRegularRing_ }

Now, in the XML export, you can just call isRegularRing():
- if true, use the existing algo section;
- if false, build numModulesPerRing Pospart sections

src/Extractor.cc
```diff
@@ -2598,6 +2651,19 @@ namespace insur {
             pconverter << myRingInfo.radiusMid;
             alg.parameters.push_back(numericParam(xml_radius, pconverter.str()));
             pconverter.str("");
+            if(myRingInfo.discNumber <=8 && (phi_two[ringIndex]).size()>0 ){
+              alg.parameters.push_back(numericParam("hasExtraInfo", "1"));
```
ghugo83, line 2655: You should not use this algorithm anyway, but just for consistency I would mention:
Please avoid calling variables **hasExtraInfo**, because after other developments are made, one forgets what the extra info is. This would be confusing if the previous algo arguments were called 'NewDev', 'isShiftedGeo', etc.
Here the idea is **isRingRegular** or similar.
But indeed not important at prototyping phase.

src/Extractor.cc
```diff
@@ -3382,6 +3448,20 @@ namespace insur {
     return res.str();
   }
 
+ std::string Extractor::arbitraryLengthVector(std::string name, std::vector<double> invec){
```
ghugo83, line 3451: - Vector copy:
Whenever you see a vector, please make sure you are not copying it by value when not needed.
Here this vector is only read: **you should pass it by const reference**.

- Variable names:
Nice to be speicfic about names: for example here, what is it the name of? You could use moduleName, ladderName, parentVolumeName, etc. 
Same for invec: ideally, the names should already tell you what the variable represent. Instead, one should not repeat the type name: it is clear that elements in a vector, are in a vector. Rather use: modulesInRingTranslations, or modulesInRingRotations, or modulesInRingPositions, etc.

src/Ring.cc
```diff
@@ -190,7 +222,7 @@ void Ring::buildTopDown() {
 
   if (numModules.state()) numMods = numModules();
   else numModules(numMods);
-  buildModules(emod, numMods, smallDelta());
+  buildModules(emod, numMods, smallDelta(),phiShift());
```
ghugo83, line 225: Same, would be nice to be more specific about what phiShift represents.

Alright, sorry for the long review. These are only suggestions, hope it helps!
Cheers

include/DetectorModule.hh
adewit: > Is this a function to extract the module's corners? I thought it already existed.

Yes, it is. I didn't find an existing function, but I'll try to improve it because it's not a very nice solution in any case.

include/DetectorModule.hh
adewit: `getRAxis()` is probably actually not needed here.

RE the naming: happy to go with whatever the consensus is here.

include/Ring.hh
adewit: Indeed - this is one of the parts that could do with improvement (both for readability of the config and for avoiding the construction 2 comments below)

geometries/CMS_Phase2/Pixel/Pixel_V7/FPIX1_7_1_0.cfg
adewit: The sensors aren't all equidistant in phi: there are 2 places around the ring where there is additional space, the exact placement is determined in Ring.cc based on the number of modules and the phiShift() - of course it could also have been passed here on a module-by-module basis, can see benefits to either :-) 

In any case, the x and y shifts are still determined as `r cos(phi)` and `r sin(phi)` (as for modules in a regular ring) - IMO it's less error-prone to let the code determine those shifts than to calculate them by hand beforehand.

src/Ring.cc
```diff
@@ -79,30 +79,37 @@ std::pair<double, int> Ring::computeOptimalRingParametersRectangle(double module
...
+    if(mod->twistAngleFromConfig() > -999){
+      mod->notInRegularRing();
       double tmp_r = mod->center().Rho();
       mod->translateR(-tmp_r);
-      mod->twist(mod->propertyTree().get<double>("twistAng"));
-     if(mod->propertyTree().get<double>("rhoCentre",0) > 0 ){
-        mod->translateR(mod->propertyTree().get<double>("rhoCentre")-mod->center().Rho());
+      mod->twist(mod->twistAngleFromConfig());
+      if(mod->rhoCentre() > 0 ){
+        mod->translateR(mod->rhoCentre()-mod->center().Rho());
```
alkemyst, line 104: Add a comment line and change the name to manualRhoCenter or something

alkemyst, line 104: Add a comment line and change the name to manualRhoCenter or something

src/Ring.cc
```diff
@@ -79,30 +79,37 @@ std::pair<double, int> Ring::computeOptimalRingParametersRectangle(double module
...
+     templ->store(moduleNode.at(i));
+    }
     EndcapModule* mod = GeometryFactory::clone(*templ);
-    if (moduleNode.count(i) > 0){ mod->store(moduleNode.at(i));}
+    mod->build();
+    mod->translate(XYZVector(modTranslateX, 0, 0));
     mod->myid(i+1);
-    std::cout<<"------begin module printout--------"<<std::endl;
-    if(mod->propertyTree().get<double>("twistAng",-99) > -99 ){//Need to do this before further determining phi/z of the module
+    if(mod->twistAngleFromConfig() > -999){
```
alkemyst, line 98: Change this into a constant somewhere (like invalid_angle)

src/Ring.cc
```diff
@@ -179,17 +179,17 @@ void Ring::buildBottomUp() {
 
   }
 
-  emod->store(propertyTree());
+  /*emod->store(propertyTree());
```
alkemyst, line 182: I see dead code....

src/Ring.cc
```diff
@@ -213,16 +213,16 @@ void Ring::buildTopDown() {
   int numMods = optimalRingParms.second;
 
   EndcapModule* emod = GeometryFactory::make<EndcapModule>(rmod, subdetectorName());
-  emod->store(propertyTree());
-  emod->build();
-  emod->translate(XYZVector(buildStartRadius() - rmod->length()/2, 0, 0));
+  //emod->store(propertyTree());
```
alkemyst, line 216: Dead code

include/DetectorModule.hh
ghugo83: Yes basically twist makes more think of the deformation of a solid (a torsion?) and is not transparent on the rotation axis.
In the private branch I had done I had used yawAngle (https://en.wikipedia.org/wiki/Yaw_(rotation)), I would suggest that, but as you want.

alkemyst:
Hi @adewit 
is this pull request still relevant?

adewit:
Let's close it for now - I need to work on the code again to take care of the new hierarchy levels (+ associated change in detID assignment), so perhaps cleaner to make a new PR when that is ready.

PR #558, opened 2021-03-05
Proudly introducing Centos8

alkemyst:
Title says it all :-)

### DIFF
--- a/src/MainConfigHandler.cc
+++ b/src/MainConfigHandler.cc
@@ -41,11 +41,11 @@ mainConfigHandler& mainConfigHandler::instance() {
 }
 
 bool mainConfigHandler::checkDirectory(string dirName) {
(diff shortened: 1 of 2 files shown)


src/MainConfigHandler.cc
```diff
@@ -41,11 +41,11 @@ mainConfigHandler& mainConfigHandler::instance() {
 }
 
 bool mainConfigHandler::checkDirectory(string dirName) {
-  if (! filesystem::exists(dirName)) {
+  if (! boost::filesystem::exists(dirName)) {
```
ghugo83, line 44: Is the source still also compatible with Centos7 after this change?

ghugo83: Ha there is `using namespace boost;` anyway

PR #563, opened 2021-03-27
Bricked export

adewit:
This adds a new bricked module type so the `isBricked` property can be written into the XML's. Nothing changes in the TkLayout analysis (no new resolutions etc.), at this point really just to make it easier to set the property in the XML's correctly. 

No configs included in the PR for now, but an example geometry can be found here: https://adewit.web.cern.ch/adewit/layouts/OT800_IT700_BRICKED/info.html

### DIFF
--- a/include/Sensor.hh
+++ b/include/Sensor.hh
@@ -53,7 +53,7 @@ public:
     stripLengthEstimate("stripLengthEstimate", parsedOnly()),
     numROCX("numROCX", parsedOnly()),
     numROCY("numROCY", parsedOnly()),
-    isBricked("isBricked", parsedOnly(),false),
+    isBricked("isBricked", parsedOnly(), false),
     sensorThickness("sensorThickness", parsedAndChecked()),
     type("sensorType", parsedOnly(), SensorType::None),
     powerPerChannel("powerPerChannel", parsedOnly()),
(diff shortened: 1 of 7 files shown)


include/Sensor.hh
```diff
@@ -52,6 +53,7 @@ public:
     stripLengthEstimate("stripLengthEstimate", parsedOnly()),
     numROCX("numROCX", parsedOnly()),
     numROCY("numROCY", parsedOnly()),
+    isBricked("isBricked", parsedOnly(),false),
```
ghugo83, line 56: spacing here

src/XMLWriter.cc
```diff
@@ -966,6 +966,9 @@ namespace insur {
       if( param.second.find("TOBDet") != std::string::npos) param.second = xml_subdet_tobdet_1;
       if( param.second.find("TIDDet") != std::string::npos) param.second = xml_subdet_tiddet;
       stream<< xml_spec_par_parameter_first << xml_tkddd_structure << xml_spec_par_parameter_second  << param.second  << xml_general_endline;
+      if( minfo.at(i).bricked ){ // Only add the line if isBricked exists
+          stream<< xml_spec_par_parameter_first << xml_bricked_name    << xml_spec_par_parameter_second << xml_true <<xml_general_endline;
```
ghugo83, line 970: and here,

otherwise looks ok :)

adewit: Ok, hope it's better now!

ghugo83:
Great, yes it does
@alkemyst ok if I merge it?

alkemyst:
Totally


---- ghugo83 wrote ----


Great, yes it does
@alkemyst<https://github.com/alkemyst> ok if I merge it?

PR #566, opened 2021-03-29
Code reformatting test

alkemyst:
This is a test: I would like to make the codebase more uniform in terms of code style, and one interesting tool to do this is `clang-format`.

I ran `clang-format` on the whole code (via a script I added, called `formatAll.sh`).

As a result, most lines will be attributed to me via `git blame`, which I wanted to avoid. This can be worked around by using a file called `.git-blame-ignore-revs` and using `git blame` with the option `--ignore-revs-file .git-blame-ignore-revs`. This can also be made automatic by using `git config blame.ignoreRevsFile .git-blame-ignore-revs`

What I just described is shown a plot here below.

The result is pretty decent, so I propose using this tool to run a global reformatting and ask developers to run clang-format (`formatAll.sh`) before sending their commits.

Opinions?
reformat


alkemyst:
The first comment is my own. I do not like some of the changes (like breaking long lines of code that instead are more readable because they all start with the same text, moving &'s and *'s from the type to the variable and other careful indenting that improved readability).
Overall it makes the code nicer to look at, but this might be a bit excessive. Your opinions?

ghugo83:
Yes personally I really don't like it. Beehhhhhh. 
You have pointed most of the downsides indeed :)

- In my opinion, the biggest drawback is that the logic of the entire codebase becomes way less clear, because the formatter does not understand it, and just goes in terms of lines lengths. The codebase feels like written by a machine not understanding the code itself.
For example, before:
image
after:
image
There is nothing after `=`, the lines starting with `+` are not aligned anymore, and the entire blocks logic is broken. 
The read is way less fluid, because way less logical and connected.
More importantly, these types of changes makes the general codebase more error-prone & more time-taking to update.

- The references and pointers syntax becomes quite painful to read:
image
It looks like we are de-referencing a pointer each time.

- The entire git history is messed up, which can be avoided locally with some extra manipulation, but still, it would not be the case for the web github repository. 
This is quite annoying: for example, I often look at a project's github web repo, and directly on the blame of a specific file (to get a first smell / possible angle of attack of a recent change which could explain a certain behavior / a bug etc). 
Now, we would see 'clang-format' instead ;p

- The only positive thing I see, is the addition of spaces after comma. But that is more of a habit to take, and worse case, is really not that a big deal ;p

For a large codebase with many developers, an auto formatting would be necessary indeed, but here, I think the benefits are really less obvious.

Why not just a simple README on code style / guidelines? Anyway, please do what you want, butt that could be better I think.

alkemyst:
Indeed I agree with you.

I first ran it and then looked at the outcome...


---- ghugo83 wrote ----


Yes personally I really don't like it. Beehhhhhh.
You have pointed most of the downsides indeed :)

  *   In my opinion, the biggest drawback is that the logic of the entire codebase becomes way less clear, because the formatter does not understand it, and just goes in terms of lines lengths. The codebase feels like written by a machine not understanding the code itself.
For example, before:
[image]<https://user-images.githubusercontent.com/11192654/112876792-38744380-90c6-11eb-89eb-760c6fb9653a.png>
after:
[image]<https://user-images.githubusercontent.com/11192654/112876870-517cf480-90c6-11eb-936c-715df2d53979.png>
There is nothing after =, the lines starting with + are not aligned anymore, and the entire blocks logic is broken.
The read is way less fluid, because way less logical and connected.
More importantly, these types of changes makes the general codebase more error-prone & more time-taking to update.

  *   The references and pointers syntax becomes quite painful to read:
[image]<https://user-images.githubusercontent.com/11192654/112877254-d49e4a80-90c6-11eb-81de-656c9f4a658f.png>
It looks like we are de-referencing a pointer each time.

  *   The entire git history is messed up, which can be avoided locally with some extra manipulation, but still, it would not be the case for the web github repository.
This is quite annoying: for example, I often look at a project's github web repo, and directly on the blame of a specific file (to get a first smell / possible angle of attack of a recent change which could explain a certain behavior / a bug etc).
Now, we would see 'clang-format' instead ;p

  *   The only positive thing I see, is the addition of spaces after comma. But that is more of a habit to take, and worse case, is really not that a big deal ;p

For a large codebase with many developers, an auto formatting would be necessary indeed, but here, I think the benefits are really less obvious.

Why not just a simple README on code style / guidelines? Anyway, please do what you want, butt that could be better I think.

PR #567, opened 2021-06-23
Latest IT cabling map flavor (notably, GBTs reshuffling in TBPX Layers 2 & 3)

ghugo83:
This PR gathers several recent changes & brainstormings which were hanging around.

- **Base the IT cabling map on the IT 701 layout as a default (instead of IT 613).** 
The relevant change here, from the cabling map point of view, is in TFPX, where the double-disk model becomes 'a la TEPX' (ie, the 2 disks in each double-disk become complementary in phi, instead of being complementary in R).
Check robustness of the code against all regressions.

-  **GBTs reshuffling in TBPX Layers 2 & 3 (to reduce number of portcards & GBTs)**
A possible (selected) portcard design, is to have a maximum of 3 LpGBTs (instead of 2 LpGBTs) per portcard.
In TBPX L2 and L3, the number of GBTs was not as small as it could be. There were 4 GBTs among modules connected to the same power chain, while only 3 GBTs would be sufficient (in terms of e-links <-> GBTs connections). 
This deliberate choice of 4 GBTs was stemming from the previous portcard design: even with 3 GBTs among modules connected to the same power chain, 2 portcards would have been necessary anyway.
The new portcard design allows to have only 1 portcard (with 3 LpGBTs) for all modules connected to the same power chain, in TBPX L2 and L3.
There is no possible optimization in TBPX L1 and L4, because there, the number of GBTs is already as low as it can be.

- **Add a parameter to control the maximum number of modules connected to each GBT.**
An attempt was to set the maximum number of modules per GBT to 5. 
However, this would add 2 GBTs per TEPX dee side (= per TEPX MFB), in TEPX Ring 4 and 5.
Since each TEPX MFB was already connected to 12 GBTs, there would be 14 GBTs per TEPX MFB instead (not an option ;p).
Hence, the only solution was to keep the number of GBTs as it is in TEPX, and hence set the maximum number of modules connected to each GBT to 6. An extra control line is added in the LpGBT design to support 6 modules.

Before PR:
https://cms-tklayout.web.cern.ch/cms-tklayout/layouts-work/repository-git-dev/OT614_200_IT613/cablingInner.html

After PR:
https://ghugo.web.cern.ch/ghugo/layouts/it_cabling/OT800_IT701_cabling/cablingInner.html
NB: Will automatically appear after the nightly builds on cms-tklayout.web.cern.ch, at repository-git-dev/OT800_IT701 tab.

@sorfanel @alkemyst @adewit

### DIFF
--- a/include/InnerCabling/GBT.hh
+++ b/include/InnerCabling/GBT.hh
@@ -24,7 +24,7 @@ class GBT : public PropertyObject, public Buildable, public Identifiable<int> {
   typedef std::vector<Module*> Container; 
 
 public:
-  GBT(PowerChain* myPowerChain, const std::string GBTId, const int myGBTIndex, const int myGBTIndexColor, const int numELinksPerModule);
+  GBT(PowerChain* myPowerChain, const std::string GBTId, const int myGBTIndexColor, const int numELinksPerModule);
 
   // MODULES CONNECTED TO THE GBT
   const Container& modules() const { return modules_; }
@@ -71,6 +71,7 @@ public:
   const int plotPowerChainColor() const { return plotPowerChainColor_; }
 
 private:
+  const int computePlotGBTIndexInPowerChain(const int myGBTIndexInPowerChain, PowerChain* myPowerChain) const;
   const int computePlotColor(const PowerChain* myPowerChain) const;
 
   Container modules_;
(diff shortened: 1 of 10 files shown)


ghugo83:
**TBPX L2:**
- Before PR:
L2
- After PR:
L2

ghugo83:
TBPX L3:
- Before PR:
L3

- After PR:
L3

ghugo83:
_- The alternation of colors is used to distinguish different power chains (warning: of course though, same color doesn't imply same power chain).
- The alternation of modules filling styles (contour, full, or dashed) is used to show the alternation of the GBTs the modules are connected to._

ghugo83:
Presentation by Stella: https://indico.cern.ch/event/949827/contributions/4073087/attachments/2129110/3603571/IT701v4.pdf

ghugo83:
NB: Was validated by Stella at https://ghugo.web.cern.ch/ghugo/layouts/it_cabling/OT800_IT701_cabling/cablingInner.html .

PR #579, opened 2022-02-23
Realistic TFPX positions

adewit:
This includes
- Realistic TFPX positions
- Export of new subdisk hierarchy level needed to accommodate those positions
- New DetID calculation in line with this hierarchy order

As the presence of a subdisk is only a construct for CMSSW, this level does not appear in the tracker that tklayout builds up, it only exists at the level of the XML extraction. 

The subdisk hierarchy is used in the extractor if the IT config has `PixelsSubDisk` (instead of `Pixels`) as the `Tracker` label, it is *not* automatically determined based on the values of `smallDelta` and `bigDelta` - this is because geometries that do not strictly require subdisks, but can accommodate them, can exist, so it is better to specify explicitly. 

To note: XMLs that include the changes that this introduces in tklayout are already in CMSSW - those were edited by hand, this PR makes the results reproducible.

@alkemyst let me know if you have comments


src/Ring.cc
```diff
@@ -79,12 +79,44 @@ std::pair<double, int> Ring::computeOptimalRingParametersRectangle(double module
 }
 
 
-void Ring::buildModules(EndcapModule* templ, int numMods, double smallDelta) {
+void Ring::buildModules(EndcapModule* templ, int numMods, double smallDelta, double phiShift, double modTranslateX) {
```
alkemyst, line 82: This is a bit longer than it used to be. Maybe a few more comments, or moving part of the code in another member function might help clarify?

adewit: I've added a few more comments, hopefully that helps. Moving part of the code to a new member function is tricky without passing around objects that probably shouldn't be - the only other option is to keep the original buildModules and have a new one that is only called in case of an irregular set of modules (that need the extra functionality). I can do that if you prefer.

PR #583, opened 2022-04-01
Add absorbed dose information

adewit:
As per the title. Currently uses an outdated TID map; PR only to merge once we have the updated map. The irradiation page looks like https://adewit.web.cern.ch/adewit/layouts/OT800_IT700/irradiation_Outer.html with these changes, so also displays the names of the maps and the integrated luminosity. Checks done:
- Replacing dose map with the fluence maps yields the exact same results as for the fluence
- Manual sanity checks of values extracted in the code compared with the values in the input map

@alkemyst I decided a draft PR was easier after all (after having to search for the diff multiple times today) - comments welcome


alkemyst:
It looks good to me.
:+1:

adewit:
@alkemyst now with the updated dose maps - example output here: https://adewit.web.cern.ch/adewit/layouts/OT800_IT700/irradiation_Pixels.html

adewit:
> Hi @adewit ,
> that is really nice, thanks!
> Could you please also add a column in the CSV file at the bottom of the page? (Identified with “Sensors irradiation file”)

Done! (example output at the link above is also updated)

adewit:
For reference, also now have the same with Dose + fluence from the 6.2.0.5 simulation: Compare: https://adewit.web.cern.ch/adewit/layouts/OT800_IT700_dose3723/irradiation_Pixels.html (current setup as in this PR) with https://adewit.web.cern.ch/adewit/layouts/OT800_IT700_dose6205/irradiation_Pixels.html (dose,fluence from more recent simulation)

I do see some nonnegligible changes in the absorbed dose, esp in FPIX inner rings, so perhaps worth considering switching to that (pending understanding what was changed)

PR #584, opened 2022-05-13
Adding some info on how irradiation is computed, and some caveats.

alkemyst:


src/ReportIrradiation.cc
```diff
@@ -137,6 +137,18 @@ void ReportIrradiation::visualizeTo(RootWSite& site) {
   mapNamesInfo = new RootWInfo("Dose and irradiation map names");
   mapNamesInfo->setValue(mapNames_);
   settingsContent.addItem(mapNamesInfo);
+  RootWInfo* notesInfo;
+  notesInfo = new RootWInfo("Note:");
```
adewit, line 141: Thanks for adding the notes (and cleaning up the bits of information I added above, it looks much nicer on the page now!) - if I may nitpick, the `xyzInfo` name and value are always separated by a colon when printed on the webpage, so I think can just write `new RootWInfo("Note")` here.

alkemyst:
Hi @adewit , this should be fixed now. I also used more Into sections to add the note n.2 and note n.3 so that the "Note" header of the paragraph will always be formatted as the first one

PR #589, opened 2022-05-23
Add dose/fluence plots

adewit:
This adds plots showing the number of modules seeing a given dose/fluence, as well as the normalised cumulative distribution (fraction of modules above a certain dose/fluence). 

Example results here: https://adewit.web.cern.ch/adewit/layouts/OT801_IT701_COPY/irradiation_Outer.html 

For the inner tracker, some of these plots are not filled because I made separate canvases for the different module spacings in the OT, and can't figure out how to not store an empty canvas (since the code is the same for IT and OT).  I won't have time to look into this in detail for the next several weeks but I don't think it is a showstopper for now since all the information should at least be there.

@alkemyst let me know what you think.

### DIFF
--- a/include/AnalyzerVisitors/IrradiationPower.hh
+++ b/include/AnalyzerVisitors/IrradiationPower.hh
@@ -43,6 +43,8 @@ class IrradiationPowerVisitor : public GeometryVisitor {
   std::map<std::string, std::vector<const DetectorModule*> > mapTypeToFluence_;
 
  public:
+  double maxDose;
+  double maxFluence;
   MultiSummaryTable sensorsPowerSummary;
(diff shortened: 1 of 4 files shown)


alkemyst:
Hi @adewit, plots are perfect, with the corrrect titles and everything.
The names of the modules are not what we use in the project "2Ssmall","2Slarge","PSsmall","PSmed", "PSlarge".
We use 2S/PS with the spacing, like in the table with the modules summary in "geometry" tab.
image

Does this naming come from the configuraiton file?

Also, I'much prefer the code to take the naming from the modules themselves, rather than having it hardcoded.

All that being said, I think we can push this to dev and we can improve on this later.

PR #590, opened 2022-10-03
Reduce Tracker2SOuterRadius by 2cm

adewit:
To avoid clash between tracker volume and BTL. Does not affect positions of tracker modules


adewit:
@alkemyst Since this is the new default in cmssw I'll merge this (in principle doesn't affect anything, but at some point someone might pick up the pixbar.xml from tklayout, so better if it is up to date)

PR #591, opened 2022-10-27
Update of OT 2S and PS modules to latest baseline geometries

Negusbuk:
# 'Measurements' on a CAD model result in the volume of a part. Direct access to mass is only possible if the model knows the material used for a give part and the corresponding material data available in the CAD system is actually correct. When import STEP files material assignment and data are not available, therefore I only measured and provided the volume.

#  stdinclude/CMS_Phase2/OuterTracker/ModuleTypes/ptPSlarger_320
- line 41: What is hybrid thickness?
- line 43: What is outerSensorExtraLength?

# stdinclude/CMS_Phase2/OuterTracker/ModuleTypes/pt2S_320
- line 37: What is hybrid thickness?

# stdinclude/CMS_Phase2/OuterTracker/Materials/ModuleMaterials/Components/MPA_and_bumps
- Please check if there really is underfill used and how much

# stdinclude/CMS_Phase2/OuterTracker/Materials/ModuleMaterials/Components/PS_Hybrid_Spacers_XY
- In the 2.6 and 4.0 mm modules there is a AlN middle-spacer around which the flex is folded. In all versions there are two corner spacers in each hybrid. Does there have to be a splitting in the config files for this? Target volume for both middle and corner spacer is 56. Middled spacer is in the central part of target 56, corner spacer at the ends of target 56.

# stdinclude/CMS_Phase2/OuterTracker/Materials/ModuleMaterials/Components/2S_Module_Supports
- It seems that the parts in this file are the inserts + screws + washers to mount the modules. If so there must be a distinction between TB2S and TEDD. Inserts look substantially different in both sub-detectors.
- A similar file doesn't exist for the PS modules. Also here we need inserts and screws for positioning and mounting. The inserts are different in TEDD, tilted and flat TBPS.

# stdinclude/CMS_Phase2/OuterTracker/Materials/ModuleMaterials/ptPS_320_XY_components
- There will by a layer of thermal interface material (TIM) between the support structure and the PS module base plate. It should be foreseen that the three different sub-detectors use different type of TIM and also different thicknesses.

# stdinclude/CMS_Phase2/OuterTracker/Materials/disk
# stdinclude/CMS_Phase2/OuterTracker/Materials/MechanicalSupports/SupportsEndcapTEDDTop_V616.cfg
# stdinclude/CMS_Phase2/OuterTracker/Materials/MechanicalSupports/SupportsEndcapTEDD1_V351.cfg
# stdinclude/CMS_Phase2/OuterTracker/Materials/MechanicalSupports/SupportsEndcapTEDD2_V366.cfg
- These files definitely need an update but they are a mystery to me...

### DIFF
--- a/config/chemical_compounds.list
+++ b/config/chemical_compounds.list
@@ -23,6 +23,9 @@ C2H4                       0.89                       C:2                  H:4
 AlN	                       3.26	                     Al:1                  N:1
 # aluminium-nitride
 
+hBN                        2.1                        B:1                  N:1
+# hexagonal Boron nitride
+
 Si3N4                      3.183                     Si:3                  N:4
 # silicon nitride
 
--- a/config/chemical_mixtures.list
+++ b/config/chemical_mixtures.list
@@ -67,6 +67,12 @@ CF                         1.91                    C:1.0
 Epoxy	                     1.16                CH2OCN:1	 
 # EX1515 from Duccio, Composite materials, 14/01/2015
 
+Epoxy_601LV                  1.15                CH2OCN:1
+# from Polytec PT technical data sheet
+
+Epoxy_TC437                  1.35                Epoxy:0.80        hBN:0.2
+# density from Polytec PT technical data sheet; BN doping ratio to be confirmed
+
 #CFRP	                     1.84	                   C:0.78            Epoxy:0.22
 # from Duccio, Composite materials, 14/01/2015
 # Density & compo ok??
--- a/config/chemical_mixtures.list
+++ b/config/chemical_mixtures.list
@@ -312,7 +312,7 @@ FPIX_PE	                 0.95	                    PE:1
 TEDD_Steel	             7.8	                 Steel:1
 
 
-
+EN_AW-7075               2.8                 Al:0.903 Zn:0.056 Mg:0.025 Cu:0.016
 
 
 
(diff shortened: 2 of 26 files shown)


alkemyst:
Hi Andreas,
  it looks like you did quite some updating of the material files! :-)
  Are these changes proposed for tkLayout to be directly imported? Do they need some kind of revision/validation?

Thanks
Stefano

Negusbuk:
Hi Stefano,
I also made changes to the structure I think it would therefore be good if my updates could be checked before merging. As mentioned in my comments I only added correct/current values for the volume of module parts and not the mass. Is it possible to simple change the unit from e.g. g to e.g. mm^3 or does one have to calculate the mass from volumes and density?
Cheers,
 Andreas

alkemyst:
Hi Andreas,
  we have some way of converting the material to mass, but at the moment the volume is not one of them.
  The material units we can give have been so far (from `src/MaterialObject:352`)
```
      {"g", GRAMS},
      {"mm", MILLIMETERS},
      {"g/m", GRAMS_METER}
```
   but since maintainability is crucial, if you tell me that we know some materials by volume, then it would be rather simple to implement a new unit (mm3). What do you think?

Thanks

Negusbuk:
Hi Stefano,
as mentioned before measuring part in CAD typically gives the volume so a new unit mm3 would really make updating values much easier. Many thanks...
Cheers,
 Andreas

alkemyst:
Hi Andreas,
  it looks I was lucky enough that the patch to add mm^3 was very simple. I created a pull request
  https://github.com/tkLayout/tkLayout/pull/592

  Would you be able to check out the new code (repo : alkemyst branch materialVolumeUnits ) and check if the added material corresponds to what it should be?

PR #593, opened 2022-11-02
Latest cabling map

adewit:
Changes to reflect the latest cabling maps. The power chains in TEPX are split across disks, hence slightly hacky implementation. This is what the latest maps look like: https://adewit.web.cern.ch/adewit/layouts/OT801_IT640_cablingtest_v2/cablingInner.html (already cross-checked by Stella).  @alkemyst let me know if this is ok for you (there are a few things to clean up, ie things I've commented, but as you may have comments on how this has been introduced more generally, I'll wait to incorporate those at the same time)

### DIFF
--- a/src/InnerCabling/GBT.cc
+++ b/src/InnerCabling/GBT.cc
@@ -24,21 +24,11 @@ void GBT::addModule(Module* m) {
 
 /*
  * Compute GBT plot style on website.
- * To distinguish different GBTs within the same power chain, alternation of fill, contour and dashed styles is used.
+ * To distinguish different GBTs within the same power chain, alternation of fill, contour, dashed, and crosshatched styles is used.
  */
 const int GBT::computePlotStyleGBTIndexInPowerChain(const int myGBTIndexInPowerChain, PowerChain* myPowerChain) const {
-  //const bool isBarrel = myPowerChain->isBarrel();
-  //const int numGBTsInPowerChain = myPowerChain->numGBTsInPowerChain();
 
-  // Usually, only 2 plot styles are needed (for example, alternation of fill and empty plot styles) 
-  // to distinguish the GBTs among a power chain.
-  // Exception: case of an odd number of GBTs within the same power chain in the barrel:
-  // a third plot style (for example, dashed) becomes necessary.
-  const int myGBTIndexInPowerChainPlotStyle = femod(myGBTIndexInPowerChain, 4);// ( (femod(numGBTsInPowerChain, 2) == 0 ) 
-                                                // 2 styles are enough (for example, full and contour)
-                                                //? femod(myGBTIndexInPowerChain, 2)
-                                                // 3 styles are necessary (for example, full, contour, and dashed)
-                                                //: femod(myGBTIndexInPowerChain, 3));
+  const int myGBTIndexInPowerChainPlotStyle = femod(myGBTIndexInPowerChain, 4);
  
   return myGBTIndexInPowerChainPlotStyle;
 }
--- a/src/InnerCabling/GBT.cc
+++ b/src/InnerCabling/GBT.cc
@@ -27,18 +27,18 @@ void GBT::addModule(Module* m) {
  * To distinguish different GBTs within the same power chain, alternation of fill, contour and dashed styles is used.
  */
 const int GBT::computePlotStyleGBTIndexInPowerChain(const int myGBTIndexInPowerChain, PowerChain* myPowerChain) const {
-  const bool isBarrel = myPowerChain->isBarrel();
-  const int numGBTsInPowerChain = myPowerChain->numGBTsInPowerChain();
+  //const bool isBarrel = myPowerChain->isBarrel();
+  //const int numGBTsInPowerChain = myPowerChain->numGBTsInPowerChain();
 
   // Usually, only 2 plot styles are needed (for example, alternation of fill and empty plot styles) 
   // to distinguish the GBTs among a power chain.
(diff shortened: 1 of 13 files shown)


src/InnerCabling/GBT.cc
```diff
@@ -27,18 +27,18 @@ void GBT::addModule(Module* m) {
  * To distinguish different GBTs within the same power chain, alternation of fill, contour and dashed styles is used.
  */
 const int GBT::computePlotStyleGBTIndexInPowerChain(const int myGBTIndexInPowerChain, PowerChain* myPowerChain) const {
-  const bool isBarrel = myPowerChain->isBarrel();
-  const int numGBTsInPowerChain = myPowerChain->numGBTsInPowerChain();
+  //const bool isBarrel = myPowerChain->isBarrel();
```
alkemyst, line 30: This should be removed

src/InnerCabling/GBT.cc
```diff
@@ -27,18 +27,18 @@ void GBT::addModule(Module* m) {
...
 
   // Usually, only 2 plot styles are needed (for example, alternation of fill and empty plot styles) 
   // to distinguish the GBTs among a power chain.
   // Exception: case of an odd number of GBTs within the same power chain in the barrel:
   // a third plot style (for example, dashed) becomes necessary.
-  const int myGBTIndexInPowerChainPlotStyle = ( (!isBarrel || femod(numGBTsInPowerChain, 2) == 0 ) 
+  const int myGBTIndexInPowerChainPlotStyle = femod(myGBTIndexInPowerChain, 4);// ( (femod(numGBTsInPowerChain, 2) == 0 ) 
                                                 // 2 styles are enough (for example, full and contour)
-                                                ? femod(myGBTIndexInPowerChain, 2)
+                                                //? femod(myGBTIndexInPowerChain, 2)
```
alkemyst, line 39: Commented-out lines should be removed here

src/InnerCabling/InnerCablingMap.cc
```diff
@@ -64,33 +64,45 @@ void InnerCablingMap::connectModulesToGBTs(std::map<int, std::unique_ptr<PowerCh
     const int layerOrRingNumber = (isBarrel ? layerNumber : ringNumber);
     const int numELinksPerModule = inner_cabling_functions::computeNumELinksPerModule(subDetectorName, layerOrRingNumber);
 
-    const std::pair<int, double> gbtsInPowerChain = computeNumGBTsInPowerChain(numELinksPerModule, numModulesInPowerChain, isBarrel);
+    std::pair<int, double> gbtsInPowerChain = computeNumGBTsInPowerChain(numELinksPerModule, numModulesInPowerChain, isBarrel);
     const int numGBTsInPowerChain = gbtsInPowerChain.first;
     const double numModulesPerGBTExact = gbtsInPowerChain.second;
     myPowerChain->setNumGBTsInPowerChain(numGBTsInPowerChain);
-
+    const bool isSplitOverRings = myPowerChain->isSplitOverRings(); //In this case we have to compute the correct lpGBT assignment in a more hacky way
```
alkemyst, line 71: Comment unclear:  does"in this case" mean that the bool isSplitOverRings will be used to decide to use the more hacky way?

src/InnerCabling/ModulesToPowerChainsConnector.cc
```diff
@@ -38,6 +38,7 @@ void ModulesToPowerChainsConnector::visit(BarrelModule& m) {
 
   const int phiUnitRef = inner_cabling_functions::computePhiUnitRef(rodPhi_, halfNumRods, isPositiveZEnd);
   const int modulePhiRefInPowerChain = femod(inner_cabling_functions::computePhiUnitRef(rodPhi_, numRods_, isPositiveZEnd), 2);
+  //const int modulePhiRefInPowerChain = femod(inner_cabling_functions::computePhiUnitRef(rodPhi_, numRods_, isPositiveZEnd), 2);
```
alkemyst, line 41: // cleanup

src/InnerCabling/ModulesToPowerChainsConnector.cc
```diff
@@ -173,6 +187,20 @@ const std::pair<int, int> ModulesToPowerChainsConnector::computeForwardModulePhi
       modulePhiRefInPowerChain = phiUnitRef - numModulesInPowerChain; 
     }
   }
+
```
alkemyst, line 190: /*
  cleanup
*/

src/InnerCabling/GBT.cc
alkemyst, line 39: cleanup comments and clarify

alkemyst (changes requested):
Hi @adewit 
I do not understand the cabling algorithm, but generally the PR seems good. There's a bit of cleanup, but besides that, I do not see any problem with this, as long as it's tested, and it works :-)

src/InnerCabling/InnerCablingMap.cc
adewit: Tried to clarify it. Indeed as you say, the bool is used to decide to use the more hacky way

src/InnerCabling/GBT.cc
adewit: I've cleaned them up - in the past an automatic decision on the number of plot styles needed (2 or 3) was made, with the new setup it's more tricky and I decided simply to always use up to 4, to avoid not being able to distinguish neighbouring areas connected to a different lpGBT

adewit:
@alkemyst thanks for the feedback and apologies for the delay in addressing it, I've cleaned up the PR now

PR #594, opened 2023-02-10
Adding IT704 - bug fix of IT702

adewit:
Wrong zOverlap was set for TBPX L2 - led to clash of dead areas (not active areas). This fixes that. 

NB *this replaces CMSSW model T25, the website still needs to be updated to reflect that*


adewit:
I assume this is OK to merge, so will do so now :-)

alkemyst:
:+1:  :-)

PR #596, opened 2023-06-21
Merging dev into master

adewit:
Probably time to do this given the many developments :)


include/DetectorModule.hh
```diff
@@ -148,11 +148,13 @@ public:
   Property<double, AutoDefault> chipNegativeXExtraWidth;
   Property<double, AutoDefault> chipPositiveXExtraWidth;
   Property<double, AutoDefault> outerSensorExtraLength;
+  Property<double, Default> centralDeadAreaLength;
 
   Property<bool, Default> removeModule;
 
   Property<double, Default> yawAngleFromConfig;
   Property<double, Default> manualRhoCentre;
+  Property<double, Default> offsetForSensors;
```
alkemyst, line 157: Hi,
it is not immediately clear from this what the offset is supposed to move the sensor by. Can we add a comment here?

adewit: Sorry for the delay - I've added a comment now. Hope that helps.

PR #599, opened 2023-11-22
Adapt tklayout to modification introduced in cms-sw/cmssw#43007 and cms-sw/cmssw#43133

Raffaella07:
cms-sw/cmssw#43007 -- pixel dead area overlap fix  -- in Extractor.cc
cms-sw/cmssw#43133 -- tracker Volume redefinition -- in include/include/tk2CMSSW_strings.hh and xml/trackerVolumeTemplate.xml


src/Hit.cc
```diff
@@ -195,8 +195,7 @@ void Hit::fillModuleLocalResolutionStats() {
 void Hit::setHitPassiveElement(const insur::InactiveElement* myPassiveElem) {
 
   if (myPassiveElem) m_hitPassiveElem = myPassiveElem;
-  else m_hitPassiveElem = nullptr;
-  //else logWARNING("Hit::setHitPassiveElement -> can't set inactive element to given hit, pointer null!");
+  else logWARNING("Hit::setHitPassiveElement -> can't set inactive element to given hit, pointer null!");
```
alkemyst, line 198: isn't maybe `m_hitPassiveElem = nullptr` missing here?

include/CoordinateOperations.hh
```diff
@@ -72,14 +73,21 @@ namespace CoordinateOperations {
     return maxget(polygon.begin(), polygon.end(), [](const XYZVector& v) { return v.Rho(); });
   }
 
-  template<class Polygon> Polygon* computeTranslatedPolygon(const Polygon& basePolygon, double normalOffset, double yOffset) {
```
alkemyst, line 75: Out of my curiosity: how else is yOffset obtained, if it's not here?

src/Extractor.cc
```diff
@@ -4749,7 +4749,7 @@ namespace insur {
 
       // Dead area Front of centre (Inactive silicon around sensor)
       if(module.numSensors()==2){
-        const double myDeadAreaFrontOfCentreWidth = modWidth + 2.* deadAreaExtraWidth;
```
alkemyst, line 4752: Why is deadAreaExtraWidth disappeared?

src/Hit.cc
bartosik-hep: I thought that `m_hitPassiveElem = nullptr` was introduced by Adinda to suppress warnings caused by an overlap between dead areas in the split-sensor geometry, which then anyway reappeared as an `ERROR` in [`Analyzer::fillRIPlotsPerMechanicalCategoryAndSubdetector`](https://github.com/tkLayout/tkLayout/blob/d5022c8ce69ff9c7bc041f27a5f351c75e855934/src/Analyzer.cc#L3752) down the chain. Now that the overlap is fixed, it's not relevant.
And the `nullptr` is anyway assigned in the `Hit` [initializer](https://github.com/tkLayout/tkLayout/blob/d5022c8ce69ff9c7bc041f27a5f351c75e855934/src/Hit.cc#L64C5-L64C21).

src/Extractor.cc
alkemyst: Ah, ok, thanks!!

PR #600, opened 2023-12-21
Dead-area material in split-sensor PXB modules

bartosik-hep:
Fixed the inconsistent amount of material in the dead-area Si strips in the split-sensor PXB modules, used in `IT711` geometry.
Previously used weight of Si, inherited from `IT704`, was calculated for 33% smaller area (150um around one big sensor plane) compared to the split-sensor geometry: 175um around two smaller sensor planes.

This affects both the material budget/weights reported by tkLayout and the material density in the exported XMLs for CMSSW.

### DIFF
--- a/src/Extractor.cc
+++ b/src/Extractor.cc
@@ -5063,12 +5063,8 @@ namespace insur {
 	      + vol[xml_PixelModuleDeadAreaFront]->getVolume()
 	      + vol[xml_PixelModuleDeadAreaBack]->getVolume();
             if (module.numSensors()==2){
-            deadAreaTotalVolume_mm3 = vol[xml_PixelModuleDeadAreaRight]->getVolume()
-	      + vol[xml_PixelModuleDeadAreaBackOfCentre]->getVolume()
-	      + vol[xml_PixelModuleDeadAreaFrontOfCentre]->getVolume()
-	      + vol[xml_PixelModuleDeadAreaLeft]->getVolume()
-	      + vol[xml_PixelModuleDeadAreaFront]->getVolume()
-	      + vol[xml_PixelModuleDeadAreaBack]->getVolume();
+              deadAreaTotalVolume_mm3 += vol[xml_PixelModuleDeadAreaFrontOfCentre]->getVolume() 
+	        + vol[xml_PixelModuleDeadAreaBackOfCentre]->getVolume();
             }
           }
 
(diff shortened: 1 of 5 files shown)


bartosik-hep:
@alkemyst could you please review and merge this PR if it's ok. This is the last step necessary to fix geometries `T33` and `T35` in CMSSW

src/Extractor.cc
```diff
@@ -5063,12 +5063,8 @@ namespace insur {
 	      + vol[xml_PixelModuleDeadAreaFront]->getVolume()
 	      + vol[xml_PixelModuleDeadAreaBack]->getVolume();
             if (module.numSensors()==2){
-            deadAreaTotalVolume_mm3 = vol[xml_PixelModuleDeadAreaRight]->getVolume()
-	      + vol[xml_PixelModuleDeadAreaBackOfCentre]->getVolume()
-	      + vol[xml_PixelModuleDeadAreaFrontOfCentre]->getVolume()
-	      + vol[xml_PixelModuleDeadAreaLeft]->getVolume()
-	      + vol[xml_PixelModuleDeadAreaFront]->getVolume()
-	      + vol[xml_PixelModuleDeadAreaBack]->getVolume();
+              deadAreaTotalVolume_mm3 += vol[xml_PixelModuleDeadAreaFrontOfCentre]->getVolume()
```
alkemyst, line 5066: why this change?

src/Extractor.cc
```diff
@@ -5113,8 +5109,8 @@ namespace insur {
       volumes.push_back(vol[xml_PixelModuleChip]);
       volumes.push_back(vol[xml_PixelModuleDeadAreaRight]);
      if(module.numSensors()==2){
-        volumes.push_back(vol[xml_PixelModuleDeadAreaBackOfCentre]);
         volumes.push_back(vol[xml_PixelModuleDeadAreaFrontOfCentre]);
```
alkemyst, line 5112: Is the order relevant here?

alkemyst (commented):
All the not-commented lines are straightforward OK

src/Extractor.cc
bartosik-hep: This is to reduce code duplication and to make the code logics easier to follow.  
Instead of calculating `deadAreaTotalVolume_mm3` from scratch in the split-sensor case I'm reusing the volume already calculated for the standard case and just adding volumes specific for this special case.

src/Extractor.cc
bartosik-hep: No, it's irrelevant, but everywhere else in the code `Front` goes before `Back`.  
Stylistically it's more consistent this way.

alkemyst (approved):

PR #634, opened 2025-06-30
Fixing minPhi and maxPhi computation in modules

alkemyst:
Some legacy code was making assumptions on the orientation and rotation of modules to compute minPhi and maxPhi

These two parameters should identify the bounding cylinder of the module in the 3D cylindrical coordinates, and should be placed between -π and 3π.

This computation **fail with significant module yaw**, like those expected in the Inner Tracker (including 180° yaw).

It turns out that these **were already failing in the TBPS tilted section**, where the number of layers with a hit was correct (because modules overlap at minimal radius), but it was failing to compute correctly whether a track could possibly hit a module in the outer radial position of the tilted module.

These values are computed prior to checking the full 3D inversion to know whether a module is hit by a track, so that modules that CANNOT be hit by a track are discarded. By comparing the new code output with the old code, we can see that more hits are found, and therefore by construction, the new code is better (fake tracks cannot be added by a bug here).

The new method was also tested to be safe against 180° rotation of the modules (`yawFlip`)

See attached plots for a direct comparison.
[minPhi_maxPhi_fixpdf.pdf](https://github.com/user-attachments/files/20980675/minPhi_maxPhi_fixpdf.pdf)


alkemyst:
This patch does not significantly affect run time.

PR #639, opened 2026-02-05
More user-friendly configuration setup

bartosik-hep:
Improved installation step with more user-friendly process of questions:

1. just press ENTER to accept default values
2. retry question if non-existent folder specified
3. minor clarifications in the README and questions

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,9 @@
 # Getting the repository
 
-     git clone https://github.com/tkLayout/tkLayout.git
-     cd tkLayout
+    git clone https://github.com/tkLayout/tkLayout.git
+    cd tkLayout
 
-The latest official version is at tkLayout/master branch, the latest development version at tkLayout/dev branch.
+The latest official version is at `tkLayout/master` branch, the latest development version at `tkLayout/dev` branch.
 
 
 # Environment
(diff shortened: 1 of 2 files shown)


src/MainConfigHandler.cc
```diff
@@ -53,64 +54,76 @@ bool mainConfigHandler::checkDirectory(string dirName) {
...
-  cout << "*** Specify the list of transverse momenta to be used for the" << endl
+  cout << "*** Q4/6. Specify the list of transverse momenta to be used for the" << endl
       << "    tracking performance test (in GeV/c)" << endl
-      << "    Example: 1, 10, 100 : ";
-  cin >> tempString;
-
+      << "    Default: " << defaultString << " : ";
+  getline(cin, tempString);
+  if (tempString.empty()) tempString = defaultString;
   getline(cin,tempString2);
```
bartosik-hep: @alkemyst Is there a specific reason for requesting user input twice for these lists of values? It's a bit confusing when nothing happens after pressing ENTER. Might be worth deleting this part as well.

bartosik-hep:
Yes, I've tested it on Alma Linux 9 and all worked fine

PR #643, opened 2026-02-10
Layer.cc, MainConfigHandler and install.sh

gabrielribcesario:
`Layer.cc`
- Initialize `zEndInner_REAL` and other variables to resolve `-Werror=maybe-uninitialized`

`MainConfigHandler`
- Improved the `mainConfigHandler::checkDirectory` logic: automatically creates missing directories when possible instead of throwing an error. If creation fails or the path is invalid, it issues a warning and re-prompts the user for input
- Added default options selection to the `MainConfigHandler` TUI: [y/n] question allows for global default options to be written to `~/.tkgeometryrc`, individual defaults can also be accepted during setup

`install.sh`
- `make install` now correctly copies all files to the TKG `standard` and `layout` directories

### DIFF
--- a/src/MainConfigHandler.cc
+++ b/src/MainConfigHandler.cc
@@ -15,28 +15,17 @@
 
 #include <sys/types.h>
 
+#include "global_constants.hh"
 #include "MainConfigHandler.hh"
 #include "Units.hh"
 
 using namespace std;
 using namespace boost;
 
-static const char* HOMEDIRECTORY = getenv("HOME");
-static constexpr char CONFIGURATIONFILENAME[] = ".tkgeometryrc";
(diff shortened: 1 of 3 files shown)


src/MainConfigHandler.cc
```diff
@@ -8,22 +8,30 @@
...
-#include <MainConfigHandler.hh>
+#include "MainConfigHandler.hh"
 #include "Units.hh"
 
 using namespace std;
 using namespace boost;
 
-template <class T> bool from_string(T& t, const std::string& s, 
-                                    std::ios_base& (*f)(std::ios_base&)) {
+static const char* HOMEDIRECTORY = getenv("HOME");
```
alkemyst: Good, but rather make them static const public class variables in MainConfigHandler

PR #644, opened 2026-02-13
Remove 'using namespace std;' from headers

gabrielribcesario:


gabrielribcesario:
Add the (new) commit hash to .git-blame-ignore-revs after the changes have been merged

PR #668, opened 2026-04-17
Extractor bug

gabrielribcesario:
The function `Extractor::analyseEndcapContainer` is supposed to operate on Z+ modules only.

- Added missing layer check in `Extractor::analyseEndcapContainer`.
- Safer `ringsIndexes` iteration.
- Removed some dead code.

### DIFF
--- a/src/Extractor.cc
+++ b/src/Extractor.cc
@@ -156,23 +156,6 @@ namespace insur {
     r.insert(std::pair<const std::string,Rotation>(rot.name,rot));
 #endif
 
-    // EndcapRot is not needed any more
-    /*
-       rot.name = xml_endcap_rot;
-       rot.thetax = 90.0;
-       rot.phix = 90.0;
-       rot.thetay = 90.0;
-       rot.phiy = 180.0;
-       rot.thetaz = 0.0;
-       rot.phiz = 0.0;
-       r.push_back(rot);
-       */
-
-    // These seem not to be needed here
-    //LogicalInfo logic;
-    //AlgoInfo alg;
-    //SpecParInfo spec;
-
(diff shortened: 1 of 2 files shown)


gabrielribcesario:
Validated with

``` bash
tklayout -T -R -w geometries/CMS_Phase2/OT806_IT742.cfg --xml
```

src/Extractor.cc
```diff
@@ -427,11 +391,13 @@ namespace insur {
     auto el = lagg.getEndcapLayers();
     int layer = 1;
     int n_of_layers = el->size();
-
-    //lagg.postVisit();   
-    //std::vector<std::vector<ModuleCap> >& ec = lagg.getEndcapCap();
     
     for (oiter = ec.begin(); oiter != ec.end(); oiter++) {
+      if (lagg.getEndcapLayers()->at(layer - 1)->minZ() <= 0) {  // skip z- endcap layers
```
alkemyst, line 396: Unclear why we look only at the positive side

src/Extractor.cc
alkemyst, line 429: This was an ugly patch by Stefano, just to avoid using unused variables, but I have no idea of why we set first for positive side only.

src/Materialway.cc
```diff
@@ -2004,22 +1784,6 @@ namespace material {
     tracker.accept(visitor);
   }
 
-  /*
-  InactiveElement* Materialway::buildOppositeInactiveElement(InactiveElement* inactiveElement) {
```
alkemyst, line 2008: There should be an equivalent of this somewhere.... where is that?

alkemyst (changes requested):

PR #671, opened 2026-05-05
Align Module Orientation and Improve XML Export Descriptions

### LINKED ISSUE #666: Fix module orientation export towards CMSSW
Use the attached presentation as reference for module orientation, and make sure both OT and IT are exported correctly.
For IT, I suggest checking the global coordinates representation of the local axes of the following four modules in TEXP (see attached presentation for details):

Image

Image

For other sections (TBPX, TBPS flat and tilted, TB2S and TEDD), use a similar approach.

[mersi_20251104_tklayout_updates.pdf](https://github.com/user-attachments/files/26672123/mersi_20251104_tklayout_updates.pdf)

gabrielribcesario:
This PR introduces refactorings to the underlying module placement logic, improves code safety, simplifies orientation calculations, and updates the TBPS v8.0.7 configuration file layout.

The main goal of these changes is to align polygon local-to-global coordinate operations, as well as align the orientation of the exported XMLs with the internal tkLayout geometry descriptions.

This fixes #666.

---

# Core Geometry & Rotations

* **Local Axes Alignment**: Reordered module transformations so that rotations around local axes are executed first. Derived local axes directly from the vertex winding sequence in `RectangularModule::build`.
* **Yaw and Orientation Fixes**: Adjusted yawed barrel modules to rotate about the local $Z$ axis instead of the local $X$ axis. Reordered local $X$ and $Y$ rotations, and isolated initial global orientation logic outside of `Barrel/EndcapModule::build`.
* **ROOT Rotations**: Replaced custom `RotateX/Y/Z` template structures inside `AbstractPolygon` with standard `ROOT::Math::Rotation3D` components (`RotationX/Y/Z`).
* **Data Types**: Converted member variables tracking tilt and skew angles in `GeometricModule` from `int` to `double`.

# Sensor Logic & Memory Safety

* **`Sensor::hitPoly` Refactor**: Simplified 3D module geometry tracking by projecting, scaling, and translating vertices along local axes, eliminating polygon math helpers. Safer memory management using `std::unique_ptr`.
* **Extrema Calculations**: Re-factored `DetectorModule::extremaWithHybrids()` to use the new orientation logic. Replaced direct base polygon access with explicit `getLocalX()`, `getLocalY()`, and `normal()` abstraction calls.
* **Shared Bounding-Box Geometry**: Extracted the hybrid-expanded bounding-box candidate-point construction into `CoordinateOperations::boundingBoxCandidates()`, shared by `extremaWithHybrids()` and `ModuleComplex`. Moved the expanded width/length/thickness formulas onto `DetectorModule` as static helpers consumed by both call sites; removed the duplicated `ModuleComplexHelpers`.
* **Line Intersections**: Updated the filter within `Polygon3d<NumSides>::isLineIntersecting` to properly register hits regardless of the polygon's normal vector's direction.

# Extractor & XML Export

## DDTrackerXYZPosAlgo Migration

All subdetectors now use per-module `DDTrackerXYZPosAlgo` placement blocks, replacing the phi-based (`DDTrackerPhiAltAlgo`) and ring-based (`DDTrackerRingAlgo`, `DDTrackerIrregularRingAlgo`) algorithms that could not represent per-module yaw variation.

* **OT Straight Barrel**: Each rod gets its own one-entry `XYZPosAlgo` block with explicit X/Y/Z positions and a registered rotation matrix. Rod center radius is computed as the average of ring-1 and ring-2 module $\rho$ per $\phi$ index, cancelling the ±`smallDelta` stagger offset. Secondary rod logical volumes are created for each distinct yaw state.
* **OT Tilted Barrel Rings**: Each module gets a one-entry `XYZPosAlgo` block; the rotation matrix is computed using an embedded 3×3 matrix engine and mapped back to spherical (`thetax`/`phix`) angles for the `DDrot` convention.
* **OT Endcap**: The pre-pass unconditionally collects X, Y, Z, and yaw for all modules. The rotation matrix is an exact algebraic expansion of $R_\phi \cdot R_\text{tilt} \cdot R_\text{flip} \cdot R_\text{yaw}$ at a fixed 90° tilt, correctly handling the local Z/Y axis sign inversions for $-Z$ and flipped modules.
* **IT Barrel**: Emits one `XYZPosAlgo` block per ladder; rotations registered via `getBarrelModuleRotation()`.
* **IT Endcap**: A pre-pass indexed by `[ring][phi]` collects position and yaw for each module; subdisk loops emit one block per module with rotations registered via `registerEndcapModuleRotation()`.

Legacy placement algorithms and their associated constants and helper functions have been removed.

## Extractor Refactoring

* **Dynamic Barrel Module Rotation**: Added the `getBarrelModuleRotation` function, which calculates a barrel module's Euler angles based on Z-axis yaw and dynamically registers uniquely named rotation tags into the XML export map.
* **Duplicated Code**: Merged duplicate OT straight-rod and flat-part rod placement loops into a single block.
* **Degenerate Service Volume Guard**: Service components with non-positive `rWidth` are now skipped with a warning instead of emitting an invalid shape; in disk service routing, `sectionMaxR` is clamped to `sectionMinR` (with an error log) when radial clearance is insufficient for the conversion station.

# Bug Fixes

* **Endcap Module Placement**: Fixed a bug where `manualRhoCentre` was ignored based on a module's yaw angle.
* **Endcap DetId ($-Z$ side)**: The $-Z$ endcap is cloned from $+Z$ by rotating each module 180° in yaw, mapping $\phi \to \pi - \phi$. `phiIdentifier` was not recalculated after this rotation, producing wrong DetId assignments in TEPX disks. Added `Ring::sortModulesByIncreasingPhi()` called from `rotateToNegativeZSide()` to re-order $-Z$ modules by their updated $\phi$ values before identifiers are assigned. Removed the incomplete $1 + ((3N/2 - \text{phiRef}) \;\%\; N)$ compensation formula from `EndcapDetIdBuilder`.
* **IT Endcap Yaw Sign**: In `Ring::buildModules`, `yaw()` and the phi-placement `rotateZ` are adjacent rotations about the global $Z$ axis, composing as $R_z(\phi + \text{yaw})$ regardless of flip state. The previous code applied $-\text{yaw}$ in the `ZPlus/Flipped` and `ZMinus/!Flipped` cases, introducing up to ~0.3 rad of yaw error for some IT Endcap modules. All four cases in `registerEndcapModuleRotation` now use $+\text{yaw}$.

# Visualization

* **Vizard**: Added floating-point rounding guards (`vizEpsilon`) to `Vizard::drawAxesAndNameXY` to suppress precision loss layout glitches.

# Configuration Files

* **TBPS v8.0.7**: Correctly mapped `yawFlipRods` parameters to flat module sections inside `TBPS_V807.cfg`. Moved `smallDelta` and `placeRadiusHint` options under the `FULL LAYER` block.

### DIFF
--- a/include/GeometricModule.hh
+++ b/include/GeometricModule.hh
@@ -38,7 +38,8 @@ class GeometricModule : public ModuleDecorable {
 protected:
   int numHits_ = 0;
   bool flipped_ = false;
-  int tiltAngle_ = 0., skewAngle_ = 0.;
+  double tiltAngle_ = 0.;
+  double skewAngle_ = 0.;
   double zrotAngle_ = 0.;
   Polygon3d<4> basePoly_;
   std::vector<XYZVector> contour_;
--- a/include/GeometricModule.hh
+++ b/include/GeometricModule.hh
@@ -76,7 +76,20 @@ public:
   double skewAngle() const { return skewAngle_; }
   double zrotAngle() const { return zrotAngle_; }
   bool flipped() const { return flipped_; }
-  bool flipped(bool newFlip) { flipped_=newFlip; return flipped_; }
+  bool flipped(bool shouldFlip) {
+    if (flipped_ != shouldFlip) {
+      // 180° rotation around the local Y axis
+      Polygon3d<4> flippedPoly;
+      flippedPoly << basePoly_.getVertex(1)
+                  << basePoly_.getVertex(0)
+                  << basePoly_.getVertex(3)
+                  << basePoly_.getVertex(2);
+      basePoly_ = flippedPoly;
+
+      flipped_ = shouldFlip;
+    }
+    return flipped_;
+  }
 
   void translate(const XYZVector& vector) { basePoly_.translate(vector); }
   void mirror(const XYZVector& vector) { basePoly_.mirror(vector); }
--- a/include/Polygon3d.hh
+++ b/include/Polygon3d.hh
@@ -211,7 +211,6 @@ bool Polygon3d<NumSides>::isLineIntersecting(const XYZVector& orig, const XYZVec
   double normOrig = this->getNormal().Dot(orig);
   double normDir = this->getNormal().Dot(dir);
   double d = this->getCenter().Dot(this->getNormal());
-  if (normDir < 1e-3) return false; // no fabs because if normDir < 0 we want to return false, as we're matching with the module in the opposite direction
   intersection = orig + (((d - normOrig)/normDir) * dir); // TODO: slightly inefficient here: we do not need to go 3D and then 2D again...
   return isPointInside(intersection);
 }
--- a/include/Polygon3d.hh
+++ b/include/Polygon3d.hh
@@ -9,6 +9,8 @@
 
 #include <TRandom.h>
 #include <Math/Vector3Dfwd.h>
+#include <Math/GenVector/RotationX.h>
+#include <Math/GenVector/RotationY.h>
 #include <Math/GenVector/RotationZ.h>
 #include <Math/GenVector/Rotation3D.h>
 #include <Math/GenVector/Transform3D.h>
@@ -22,44 +24,8 @@ struct PolygonLess {
   }
 };
 
-template<class Coords, class FloatType>
-struct RotateX {};
-
-template<class FloatType>
-struct RotateX<XYZVector, FloatType> {
-  XYZVector operator()(const XYZVector& vector, FloatType angle) const {
-    return XYZVector(vector.X(),
-                     vector.Y()*cos(angle) - vector.Z()*sin(angle),
-                     vector.Y()*sin(angle) + vector.Z()*cos(angle));
-  }
-};
-
-template<class Coords, class FloatType>
-struct RotateY {};
-
-template<class FloatType>
-struct RotateY<XYZVector, FloatType> {
-  XYZVector operator()(const XYZVector& vector, FloatType angle) const {
-    return XYZVector(vector.Z()*sin(angle) + vector.X()*cos(angle),
-                     vector.Y(),
-                     vector.Z()*cos(angle) - vector.X()*sin(angle));
-  }
-};
-
-template<class Coords, class FloatType>
-struct RotateZ {};
-
-template<class FloatType>
-struct RotateZ<XYZVector, FloatType> {
-  XYZVector operator()(const XYZVector& vector, FloatType angle) const {
-    return XYZVector(vector.X()*cos(angle) - vector.Y()*sin(angle),
-                     vector.X()*sin(angle) + vector.Y()*cos(angle),
-                     vector.Z());
-  }
-};
-
+// Primary templates (required before specialization)
 template<class Coords> Coords crossProduct(const Coords& v1, const Coords& v2);
-
 template<class Coords> Coords unitVector(const Coords& vector);
 
 template<int NumSides, class Coords, class Random, class FloatType = double>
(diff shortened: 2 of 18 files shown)


include/GeometricModule.hh
```diff
@@ -76,7 +76,20 @@ public:
   double skewAngle() const { return skewAngle_; }
   double zrotAngle() const { return zrotAngle_; }
   bool flipped() const { return flipped_; }
-  bool flipped(bool newFlip) { flipped_=newFlip; return flipped_; }
+  bool flipped(bool shouldFlip) {
```
alkemyst, line 80: I wonder who is using this.
I mean: doFlip could be a good method (takes the current module and flips it). And this is actually what your implementation does. If called twice, the module would go back to the original position.
But then you also keep the flag whether it was flipped. If flipped twice, it will be wrong...

alkemyst: I would not want the flipped_ value to depend on whether this function is called or not.
I suspect we should never read flipped_
do we? where? for what purpose?

include/Polygon3d.hh
```diff
@@ -34,16 +32,15 @@ protected:
...
   virtual void computeProperties() = 0;
   mutable Coords center_, normal_;
   mutable bool centerDirty_, normalDirty_;
   void setGeomDirty(bool dirty) { centerDirty_ = normalDirty_ = dirty; }
 
 public:
-  AbstractPolygon() : allocated_(0), rotation_(), centerDirty_(true), normalDirty_(true) {}
-  AbstractPolygon(const Coords& vertex): allocated_(0), rotation_(), centerDirty_(true), normalDirty_(true) { *this << vertex; }
+  AbstractPolygon() : allocated_(0), centerDirty_(true), normalDirty_(true) {}
+  AbstractPolygon(const Coords& vertex): allocated_(0), centerDirty_(true), normalDirty_(true) { *this << vertex; }
```
alkemyst, line 43: what does this mean?

include/Polygon3d.hh
```diff
@@ -127,26 +123,27 @@ AbstractPolygon<NumSides, Coords, Random, FloatType>& AbstractPolygon<NumSides,
 }
 
 template<int NumSides, class Coords, class Random, class FloatType>
-AbstractPolygon<NumSides, Coords, Random, FloatType>& AbstractPolygon<NumSides, Coords, Random, FloatType>::rotate(const ROOT::Math::Rotation3D& rot) {
-  std::transform(&v_[0], &v_[NumSides], &v_[0], [&](const Coords& coord) { return rot(coord); });
-  rotation_ = rot * rotation_; // Track total accumulation
+AbstractPolygon<NumSides, Coords, Random, FloatType>& AbstractPolygon<NumSides, Coords, Random, FloatType>::rotateX(FloatType angle) {
```
alkemyst, line 126: these two look very different. it's hard for me to tell what you are trying to do here

src/DetectorModule.cc
```diff
@@ -874,16 +874,15 @@ const int DetectorModule::innerDTCPlotColor() const {
...
-    decorated().rotateY(M_PI/2);
 
-    rAxis_ = normal();
+    rAxis_ = XYZVector(1., 0., 0.);
+
+    // Initial global orientation with skew
+    rotateY(M_PI_2 - skewAngle());
 
-    // skew
-    decorated().rotateZ(skewAngle());
```
alkemyst, line 883: The old and new look VERY different. Why?
I used to have an rAxis that went orthogonal to the module, and now we assume that that's parallel to x (?!)

And I used to have a rotation around Y by π/2 followed by a rotation around Z by the skew angle, and now it's a single rotation around Y by π/2-skew.
(?!)

src/DetectorModule.cc
```diff
@@ -895,11 +894,15 @@ void BarrelModule::build() {
 void EndcapModule::build() {
   try {
     DetectorModule::build();
-    //myModuleCap_->setCategory(MaterialProperties::e_mod);
-    rAxis_ = (basePoly().getVertex(0) + basePoly().getVertex(3)).Unit();
+
+    // Initial global position
+    rotateZ(M_PI_2);
+
+    rAxis_ = XYZVector(1., 0., 0.);
```
alkemyst: same as above. this used to be the center of the module, but unitary, and now we have hard-coded thhe x axis

alkemyst (commented):
Partial review , up to Ring.cc (from top down)

include/Polygon3d.hh
gabrielribcesario: This isn't visible if you look at the changes as a whole. It was a helper member added by the 1st commit, removed in the last.

include/GeometricModule.hh
gabrielribcesario: > I suspect we should never read flipped_

`flipped_` is read in different points of [Extractor::analyseLayers](https://github.com/tkLayout/tkLayout/blob/c799898b39b8b44c922c68b1b4aeaa462f7b5bf8/src/Extractor.cc#L548) and [analyseDiscs](https://github.com/tkLayout/tkLayout/blob/main/src/Extractor.cc#L2134) .

> I would not want the flipped_ value to depend on whether this function is called or not.

`flipped_` is automatically assigned `false` during declaration ([GeometricModule.hh](https://github.com/tkLayout/tkLayout/blob/c799898b39b8b44c922c68b1b4aeaa462f7b5bf8/include/GeometricModule.hh#L40)).

> I mean: doFlip could be a good method (takes the current module and flips it).

I also thought of renaming the method, but then it would diff a bunch of different files, which I'd rather do in a separate commit.

> If called twice, the module would go back to the original position.
But then you also keep the flag whether it was flipped. If flipped twice, it will be wrong...

The function should still work while being called twice. It still handles the "flip" command just like before:
- Before: If `newFlip` matches `flipped_`, nothing happens. If they're different, `flipped_` gets updated.
- Now: Same logic, except now whenever `shouldFlip` doesn't match `flipped_`, it also swaps the polygon's vertices (and clears the sensor poly).

